PR #1, opened 2016-06-29
Make Result a structure instead of class

yakimovim:
I suggest to make the Result to be a structure instead of class. In this case there is no really a way to return null where Result is expected.


vkhorikov:
Great work! Fantastic workaround for structs being non-inheritable!
Will publish a new package version soon.

yakimovim:
Thank you!

PR #3, opened 2016-07-06
Private error messages. Methods for conversion of exceptions to result.

yakimovim:
Frequently there are two types of error messages:

* Public error messages that should be displayed to user.
* Private error messages that should be stored into logs and analyzed by developers.

The simplest example of this separation comes from exceptions. Message of exception sometimes can be shown to user. But exception also contains stack trace which should not be displayed to user but has a great value for developer.

This pull request adds support of private error messages into Result. If private error message is not set explicitly then it will be the same as public error message.

Also several methods for creation of Result from exceptions are added.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/CombineMethodTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/CombineMethodTests.cs
@@ -35,13 +35,14 @@ namespace CSharpFunctionalExtensions.Tests.ResultTests
         public void Combine_combines_all_errors_together()
         {
             Result result1 = Result.Ok();
-            Result result2 = Result.Fail("Failure 1");
-            Result result3 = Result.Fail("Failure 2");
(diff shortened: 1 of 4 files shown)


vkhorikov:
Thank you for the pull request. I have some trouble understanding the use cases for such a feature, through.

The idea behind the Result struct is that it should contain the domain knowledge regarding the failure happened, not technical info. Also, one of the guidelines I advocate everyone follows is catching  exceptions on the lowest level possible if they come from a 3rd party library and you except them to happen or on the highest level possible for logging purposes. (More info: [blog post](http://enterprisecraftsmanship.com/2015/02/26/exceptions-for-flow-control-in-c/), [PS course](https://app.pluralsight.com/library/courses/csharp-applying-functional-principles/table-of-contents), "Refactoring Away from Exceptions" module.) The technical info can be stored to logs where the exception is caught, what is the point in putting them into a Result instance?

yakimovim:
Thank you for the explanation, Vladimir. No I see how "private" information should be processed.

PR #10, opened 2016-09-13
Add Func<> equivalent to Execute Action<> Maybe extension

laurence79:
Great library, love the approach and your Enterprise Craftsmanship series 😊 

When using the library in my own project I found that I sometimes wanted to return a value from the Maybe<>.Execute() extension method, so I've added a Func<> overload in this PR in the hope that it might be of use to others.

### DIFF
--- a/CSharpFunctionalExtensions/MaybeExtensions.cs
+++ b/CSharpFunctionalExtensions/MaybeExtensions.cs
@@ -69,5 +69,14 @@ namespace CSharpFunctionalExtensions
 
             action(maybe.Value);
         }
+
+        public static TOut Execute<T, TOut>(this Maybe<T> maybe, Func<T, TOut> func)
+            where T : class
+        {
+            if (maybe.HasNoValue)
+                return default(TOut);
+
+            return func(maybe.Value);
+        }
     }
 }
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
@@ -156,6 +156,15 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             property.Should().Be("Some value");
         }
 
+        [Fact]
+        public void Execute_returns_func_result()
+        {
+            Maybe<MyClass> maybe = new MyClass { Property = "Some value" };
+
+            string property = maybe.Execute(x => x.Property);
+
+            property.Should().Be("Some value");
+        }
 
         private static Maybe<string> GetPropertyIfExists(MyClass myClass)
         {


laurence79:
Alternatively, the following is more comprehensive but extends Nullable<<T>> as well. I didn't know if you wanted to go that far...
```
public static Maybe<TOut> Execute<TIn, TOut>(this Maybe<TIn> maybe, Func<TIn, Maybe<TOut>> func)
    where TIn:class
    where TOut:class
{
    if (maybe.HasValue)
        return func(maybe.Value);
    else
        return null;
}

public static TOut? Execute<TIn, TOut>(this Maybe<TIn> maybe, Func<TIn, TOut?> func)
    where TIn : class
    where TOut : struct
{
    if (maybe.HasValue)
        return func(maybe.Value);
    else
        return null;
}

public static Maybe<TOut> Execute<TIn, TOut>(this TIn? nullable, Func<TIn, Maybe<TOut>> func)
    where TIn : struct
    where TOut : class
{
    if (nullable.HasValue)
        return func(nullable.Value);
    else
        return null;
}

public static TOut? Execute<TIn, TOut>(this TIn? nullable, Func<TIn, TOut?> func)
    where TIn : struct
    where TOut : struct
{
    if (nullable.HasValue)
        return func(nullable.Value);
    else
        return null;
}
```

vkhorikov:
Thanks for the contribution!

Did you consider the Select extension method? It seems like it does exactly what you need (transforms a `Maybe<T1>` to a `Maybe<T2>`).

I see now that the naming I've chosen is probably confusing. Do you think both `Execute` and `Select` should be named the same way? If so, which name should we pick?

Regarding extensions for structs - I like the idea!

laurence79:
I did completely miss that 😞 I was looking for something similar to OnSuccess() for Result<>. I think Select is the logical choice, I'll work with that for now.

vkhorikov:
Still great suggestion regarding structs! I'll add it soon and will roll out a new version of the lib.

laurence79:
As overloads to Select? I'm happy to amend the PR if that helps?

vkhorikov:
>As overloads to Select?
Yes.

>I'm happy to amend the PR if that helps?
Would be very helpful, thanks!

laurence79:
So I've run into a blocker here.

The methods
```
public static Maybe<K> Select<T, K>(this Maybe<T> maybe, Func<T, K> selector)
    where T : class
    where K : class
```
and
```
public static K? Select<T, K>(this Maybe<T> maybe, Func<T, K> selector)
    where T : class
    where K : struct
```
have the same signature. Despite the different type constraints.  More info here https://blogs.msdn.microsoft.com/ericlippert/2009/12/10/constraints-are-not-part-of-the-signature/

The solution to this isn't immediately apparent to me 😕 

(My original code had Func<T, K?> selector, but of course this isn't right - the func itself should not return a nullable)

vkhorikov:
Ah, indeed, return values also don't count as overloads.
Maybe we can rename `Select` for structs to some other method? How about `SelectStruct`?

ahryshchanka:
@vkhorikov @laurence79 I can't find PR, but I have naming question:
Why in this thread discussion about `.Select(...)` method, but in library we usually name methods `.Map(...)`?

CSharpFunctionalExtensions/MaybeExtensions.cs
```diff
@@ -69,5 +69,14 @@ public static void Execute<T>(this Maybe<T> maybe, Action<T> action)
 
             action(maybe.Value);
         }
+
+        public static TOut Execute<T, TOut>(this Maybe<T> maybe, Func<T, TOut> func)
```
ahryshchanka: Looks like `Map` method, not like `Execute` or `Select`

vkhorikov: `Select` == `Map` in C#. Found some conflicts merging this PR. Could you re-submit?

SuperJMN: @a-z-hub Ping! 
Is this still active?

SuperJMN:
@laurence79 Is this PR still active? Please, fix the merge conflicts or say something :)

PR #12, opened 2016-11-20
targeting net standard 1.0

omaraloraini:
Created a new solution copied all the classes from the old one.

To create a nuget package run "dotnet build" then "dotnet pack" inside 
CSharpFunctionalExtensions project.

there is weird error that might happen about "project.lock.json" in the tests project
from project.json in the tests project delete
"CSharpFunctionalExtensions": {
          "target": "project"
},
then save, wait for the restore then insert it again and save.


vkhorikov:
Wow, thanks for such a hard work converting the library! I'll try to release a new nuget package asap.

vkhorikov:
Just pushed a new version of the lib. Thanks again!

PR #21, opened 2017-03-08
Added flatmap methods, unit tests, and examples

solvingJ:
I think this is a major fundamental addition to the library, which is present in most functional languages for constructs like Optional, Result, Either, Etc. 

I see attempts to work around the lack of FlatMap with this kind of technique, which I also used for a time:
```
.OnSuccess(customer => gateway.SendPromotionNotification(customer.Email))
.OnBoth(result => result.IsSuccess ? "Ok" : result.Error);
```

Taken from here:
https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions.Examples/ResultExtensions/AsyncUsageExamples.cs

Beyond just being more concise, there are significant challenges with the previous workaround did not solve, and FlatMap does.

### DIFF
--- a/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
+++ b/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
@@ -40,6 +40,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Properties\AssemblyInfo.cs" />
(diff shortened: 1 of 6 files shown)


solvingJ:
vlad explained that instead of Map , one should use OnSuccess to transform the a value of type TValue to a type TResult under the special case that TResult is actually of type Result. However Map should be used if TResult is type other than Result.

PR #26, opened 2017-04-07
Added Exception overload

peterdeme:
Hi,

An actual Exception was something that I missed from this library. Let me know your thoughts.

Thanks
Peter

### DIFF
--- a/CSharpFunctionalExtensions.Tests/CSharpFunctionalExtensions.Tests.csproj
+++ b/CSharpFunctionalExtensions.Tests/CSharpFunctionalExtensions.Tests.csproj
@@ -8,7 +8,7 @@
     <PackageReference Include="FluentAssertions" Version="4.19.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="xunit" Version="2.2.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta1-build1309" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


vkhorikov:
Hi Peter,

Could you provide some usage examples? When would you need a full exception as opposed to just the text from it?

Thank you,
Vladimir

peterdeme:
Hi Vladimir,

Actually, for me it's rarer to provide just a simple message, than passing an exception. I think this is a legit use case:

```csharp
public static Result<T> Deserialize<T>(Stream stream)
{
       var xmlSerializer = new XmlSerializer(typeof(T));
       try
       {				
	    return Result.Ok<T>((T)xmlSerializer.Deserialize(stream));
	}
	catch (Exception ex)
	{
	    return Result.Fail<T>(ex);
	}           
}
```
Of course, you could do ex.ToString(), but later on you can not actually investigate the exception.

One of our project has exception handler middleware, which works like that:

```csharp
if (exception is ApplicationException)
{
	//return InternalSeverError
}
else if (exception is BusinessLoginException)
{
	//return conflict
}
else if (exception is CommonWhateverException)
{
			
	//return bad request
}
//return whatever..
```

Do you see my point?
Anyways, I would like to refactor my commit a little bit, I just dont have time currently. If you are okay with the idea, give me few days please till I commit a new one.

Thanks,
Peter

vkhorikov:
Hi Peter,

Sorry for the late reply. And thanks for the elaboration, I see your point now.

I would recommend against such use of exceptions, though. In general, it's a good idea to throw exceptions in case of a 500 error, not return them as method return values. This would signalize an unrecoverable situation in code.

Similarly, if there's a validation error, I don't think using an exception to describe it would be a good choice. It'd probably be better to return a message (or some custom object) instead.

Here I wrote more on this topic:
http://enterprisecraftsmanship.com/2017/01/31/rest-api-response-codes-400-vs-500/
http://enterprisecraftsmanship.com/2017/03/13/error-handling-exception-or-result/

Thank you,
Vladimir

PR #29, opened 2017-05-04
Revert "Update to netstandard 1.3, added ISerializable"

solvingJ:
Waiting 6 days until netstandard 2.0 comes out to add tests.  Netstandard 2.0 contains the necessary BinaryFormatter not currently available in netstandard 1.x

### DIFF
--- a/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
+++ b/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.0</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


vkhorikov:
It seems that there's no need to serialize both IsFailure and IsSuccess. Could you adjust the code so that only one of them gets serialized?

solvingJ:
I actually thought about it both ways and still see both sides. One could make same argument for existence of both vars in the original struct. Why have both? its just sugar with very little value. 

I actually decided I DO like having both in the struct, but that wasn't why I included both in serialization. 

Ultimately I realized that the serialized version should include all non-null members of the original for consistency and predictablility.  Choosing to eliminate one actually seems l arbitrary after thinking about it for a bit.

 Also, without both, when deserializing, one would have to go through the troubles of polymorphic deserializing, choosing between separate "SuccessResult" and FailedResult" objects , which again aren't consistent with the original struct.

vkhorikov:
Merged the pull request. Could you add some tests for this functionality?

solvingJ:
Yes, I believe Netcore and Netstandard will be out on May 10 which will bring BinaryFormatter. This will give an effective way to test serialization without bringing in dependencies. If there's a different way to test serialization now, I couldn't find it :(

PR #30, opened 2017-05-05
Added continueOnCapturedContext parameter to Async extensions

jonny-novikov:
Explained here https://arbel.net/2014/09/21/configureawait-or-not/

### DIFF
--- a/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
+++ b/CSharpFunctionalExtensions.Examples/CSharpFunctionalExtensions.Examples.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.0</TargetFramework>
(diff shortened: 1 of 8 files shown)


jonny-novikov:
@solvingJ Errm... sorry, I cloned first the fork. But I see

vkhorikov:
Ah, it looks like the new default parameter broke one of the usage examples (Promote_with_async_methods_in_the_middle_of_the_chain), have to revert it because of that.
BTW, have you ever encountered a need in `.ConfigureAwait(true)` in practice ? That could be the case if you have to preserve the synchronization context for some reason but I haven't seen that yet (at least in server-side apps).

jonny-novikov:
@vkhorikov General case of capturing the context is not for server-side applications - it can be useful on clients (desktop, etc.)

vkhorikov:
That's true, it can be useful in client apps. But still, need to come up with a solution that doesn't break the current usages. Do you have any ideas on that?

jonny-novikov:
@vkhorikov Yeap, I'll try to return to the code on the weekdays.

PR #43, opened 2017-11-24
Multi-target support: added netstandard2.0 and net45

mjharwood:
Adding the additional targets just worked. :-)


vkhorikov:
Hey, so why did you close it?

mjharwood:
It built multiple copies, but they were all tagged as 4.5 and it needs more testing than I initially thought.

vkhorikov:
Ah, I see. Feel free to resubmit it, I will merge and republish the package.

PR #44, opened 2017-11-26
multiframework support: .net standard 2.0, .net standard 1.3, .net 4, .net 4.5, .net 4.6.1

mjharwood:
* default to Standard 2.0
* build netstandard2.0;netstandard1.3;net40;net45;net461
* added framework specific references where necessary
* labeled per-framework dll with which framework it uses
* added per-framework link tests to verify the package works with each version


vkhorikov:
Thanks, great job! Just pushed a new version to NuGet.

PR #54, opened 2018-03-02
GetObjectData logic can be moved to ResultCommonLogic

SergAtGitHub:
Hello Vladimir, 

I mentioned that you're using a `ResultCommonLogic` class and I though it would be a good idea to move a serialization logic there. There were also added several simple tests, just to check, that serialization works as expected after changes. 

Regards


vkhorikov:
Merged, thanks!

PR #55, opened 2018-03-02
New result structure with error object.

SergAtGitHub:
Hello Vladimir,

In my opinion error may be represented not only with a string, but also with a context object. I would suggest to introduce a `Result<TValue, TError>` containing not only typed value, but also a typed object.

Regards


vkhorikov:
Great idea, but the changes conflict with your other commit, had to revert them. Could you please rebase them? Also, I think it'd be a good idea to inherit `ResultCommonLogic` from `ResultCommonLogic<string>`.

Issue #53, opened 2018-03-02
default ConfigureAwait(false) is causing to lose thread culture


vukasinspasic:
I have a heavy usage of your functional library, it is doing a great job as part of API project. The problem I noticed is that after calling OnSuccess or ToResult on async methods current thread culture is lost because you are using ConfigureAwait(false) as default which doesn't return to thread caller, but to first available thread. I know OnSuccess and ToResult have parameter bool continueOnCapturedContext = false that I can set to true, but I wouldn't want to do this on every place in code. So my question actually is, why do you have ConfigureAwait(false) in your library when .NET is continuing on captured context ( ConfigureAwait(true) ) by default? And is it possible that you set continueOnCapturedContext  to have default value = true?

vkhorikov:
The primary reason was to help avoid potential dead locks where people are using methods from the library in a blocking context, such as `DoSomething().Result` where `DoSomething` returns a `Task`.

I'm actually OK with changing this behavior as people shouldn't be using `.Result` anyway and it's quite easy to avoid using it in scenarios with API.

If you could do a PR, I'd release a new version of the lib.

vukasinspasic:
Thank you very much for fast answer! I will try to make PR soon

vkhorikov:
Available in 1.10.0

vukasinspasic:
Thank you @vkhorikov and @robosek

PR #57, opened 2018-03-18
Change default ConfigureAwait to true

robosek:
I'm the new user of this library and I think it's really great! Big **Thank You** for introducing such amazing ROP library for C#.  I decided to prepare fix for #53 . I also think it's easier for us to not explicitly put `true` on every `async` extension method. So, please review it if you have time.


vkhorikov:
Thanks much! Merged it, will update the lib on NuGet later this week.

vkhorikov:
Published 1.10.0 version. Thanks again

robosek:
OK thank you for information

PR #68, opened 2018-09-26
Fix Maybe.HasValue for None on structs

### LINKED ISSUE #67: Maybe<int>.None.HasValue is true
While tracking a bug if found that Maybe<int>.None.HasValue was true whereas my expected behavior is that any None would have no values.

freakingawesome:
Fixes #67


vkhorikov:
Merged, thanks! Will publish a new version next week.

PR #71, opened 2018-10-26
Added basic overloads for Result<T, TError>

hankovich:
Hi Vladimir, thank you for this awesome lib!

**Motivation**

We widely use it in our company but we faced problems with Result<T, TError> support. It's very poor now and we wrote our own extensions for our extensions :) that are dealing with Result<T, TError>. But since we have a lot of solutions (microservices architecture and such kind of staff) we should duplicate that code in all our solution. 

So I decided to create this PR to improve Result<T, TError> support.

**What is done:**

Added Maybe<T>.ToResult overload with custom error type (which produces Result<T, TError>).
Added basic overloads for Result<T, TError>
Improved ConfigureAwait where it was missed
Added 2 tests (for Maybe, because as I see we are not testing result extensions)

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
@@ -71,6 +71,30 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


vkhorikov:
Merged, thanks! Will publish a new version this week.

vkhorikov:
Published, v1.13.0

PR #73, opened 2018-12-01
Replace reference to Microsoft.CompilerServices.AsyncTargetingPack

CarlHA:
Microsoft.CompilerServices.AsyncTargetingPack has been unlisted on nuget and has been replaced by Microsoft.Bcl.Async.


vkhorikov:
Looks like with this modification LinkTest.4.0 (the lib that checks the package works on .NET 4.0) doesn't work anymore.

PR #74, opened 2018-12-03
Added a Net4.0 project to allow use of Microsoft.Bcl.Async

CarlHA:
Resolves issue with LinkTest4.0.  Due to the issues with Microsoft.Bcl.Async in a multi target project, a separate .Net4.0 project has been created.  This will need to be released to nuget as a separate package.

I have also bumped the version to 1.13.1


vkhorikov:
Is this PR implies creation of a new nuget package for .NET 4.0? If so, I would rather prefer to keep a dependency on the unlisted nuget package instead.

CarlHA:
Then in which case it will not install into a .Net 4.0 project which is why I created this PR.  When a package is unlisted, a new install will not find this dependency.

vkhorikov:
Apology for the delay. Published as a separate nuget package here: https://www.nuget.org/packages/CSharpFunctionalExtensionsNet4.0/

CarlHA:
Thanks

PR #75, opened 2018-12-08
Add OnFailureCompensate extension methods with tests and example

golavr:
I'v been playing with this library for few days after watching your pluralshight course which is awesome BTW.
I noticed that something is missing, w/o it i found myself having to fallback to imperative style within OnBoth to achieve what i needed.
I don't know if it's part of the ROP(Railway oriented programming) approach or not but i tried to find material on web... i didn't find.
In ROP you have two track model, success and failure.
They are connected by blocks(functions), in each function you can switch from success track to failure track.
Once you are on the failure track your journey is ended and only OnFailure or OnBoth chains are going to be executed.
What i'm trying to do is to switch back from the failure track to success track, i called it compensate(OnFailureCompensate).
By adding compensate i was able to stick with the functional style.

[gist example](https://gist.github.com/golavr/5f0f55da6a3b60dea1e365d3420bac9b)


vkhorikov:
Merged and published on Nuget. Thanks.

PR #78, opened 2018-12-18
Implemented deconstruction for all kinds of Result

### LINKED ISSUE #76: Result desconstruction
I want to propose the feature of deconstructing `Result`, so we can do
```
var (isSuccess, isFailure, error) = Result.Ok();
var (isSuccess, isFailure, somethingIsDisabled, error) = Result.Ok(true);
```
Motivation for me is the case when you have a primitive value, like `bool` wrapped in `Result` and we want to leave the monad e.g.
```
var serviceIsDisabled = CheckIfDisabled()

if (serviceIsDisabled.Value) // Do something
```

`serviceIsDisabled.Value` in condition looks kinda ugly to me, so i've implemented deconstruction via extension methods. It would be nice to have it in the main package.

Please, let me know if you're ok with and i'll prepare PR with tests and stuff

saitonakamura:
Fixes #76

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/DeconstructionTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/DeconstructionTests.cs
@@ -1,0 +1,167 @@
+﻿using System;
+using FluentAssertions;
+using Xunit;
+
+namespace CSharpFunctionalExtensions.Tests.ResultTests
+{
+    public class DeconstructionTests
+    {
+        [Fact]
+        public void Can_deconstruct_non_generic_Ok_to_isSuccess_and_isFailure()
+        {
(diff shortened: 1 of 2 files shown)


vkhorikov:
Merged, thanks. Although, why deconstruction to both `isSuccess` and `isFailure`? Wouldn't `isSuccess` be enough?

saitonakamura:
I find `if (isFailure)` better `if (!isSuccess)` most of the time. Usually, in imperative code when working with booleans, it's too tedious to create whole new variable just to achieve `if (failure)` semantics, but here it's for free, so why not use it.

vkhorikov:
Pushed a new version to nuget, should be there within the next couple hours. Thanks again for the contribution.

saitonakamura:
Cool, thanks!

Issue #79, opened 2018-12-26
Add extension method Combine for IEnumerable<Result<T>> / IEnumerable<Result>


ahryshchanka:
In case OnSuccess return IEnumerable<Result<T>> required to use method:

```
var result = Result.Combine(results);
return result.IsSuccess
         ? Result.Ok(results.Select(e => e.Value))
         : Result.Fail<IEnumerable<T>>(result.Error);
```

and to do it in OnSuccess method:
`.OnSuccess(results => /*code above*/)`

It will be nice to have extension methods:
`Result<IEnumerable<T>> Combine(this IEnumerable<Result<T>> results)`
`Result Combine(this IEnumerable<Result> results)`

In this case we will have something like:
```
.OnSuccess(arg => ... ) /* returns IEnumerable<Result<T>> */
.Combine() /* convert IEnumerable<Result<T>> to Result<IEnumerable<T>> */
.OnSuccess(collection => ... )
```

vkhorikov:
Looks like a good one, please go ahead with a PR.

Issue #66, opened 2018-09-05
Make error message separator , configureable


altmann:
We use this library in our project but we have one suggestion. It would be cool if it is able to configure the error message separator which is now the constant ", " on one central point. 

For example:

    Result.SetupErrorMessageSeparator("$$");

I know, it is able to pass the error message separator on method level like 

    Result.Combine("$$", r1, r2);

But this is not a solution for us because then we have to pass the separator on each usage of combine. 

Do you see a opportunity to implement this?

vkhorikov:
Don't see why not. You can create a PR and I'll publish a new version. Alternatively, I can do that myself but but that can take awhile.

altmann:
Ok, that seems fair. We will send you an PR in the next weeks/months.

altmann:
@vkhorikov Can you please publish a new version to Nuget. We need this new feature in the next days. 

Thank you...

vkhorikov:
Published: 1.17.0

PR #80, opened 2018-12-26
Add Combine extension methods (#79)

ahryshchanka:


hankovich:
Hi, I understand that you added only mentioned overloads, but I think we need to have more overloads.
What do you think about:

1. IEnumerable<Task<Result>> / IEnumerable<Task<Result<T>>>
2. Task<IEnumerable<Result>> / Task<IEnumerable<Result<T>>>
3. Task<IEnumerable<Task<Result>> /  Task<IEnumerable<Task<Result<T>>> (Wow)
4. What should we do with Result<T, TError>?
5. Sometimes in case of IEnumerable<T> we don't want to see Result<IEnumerable<T>>, but we want to have Result<IEnumerable<TNew>>, so
```public static Result<IEnumerable<TNew>> Combine<T, TNew>(this IEnumerable<Result<T>> results, Func<T, TNew> mapper```
6. We should also have ability to compose resulted IEnumerable to a single object (like `SelectMany` does)
```public static Result<TResult> Combine<TValue, TResult>(this IEnumerable<Result<TValue>> results, Func<IEnumerable<TValue>, TResult> composer)```
7. Overloads without errorMessagesSeparator should be added as well (You should use ", " as a default separator)

**Comment by grishchenkoa** [comment]
`async Task<Result> Combine(this IEnumerable<Task<Result>> tasks, string errorMessageSeparator, bool continueOnCapturedContext = true)` method added

**Comment by grishchenkoa** [comment]
@vkhorikov and @hankovich I think I did all required changes and code ready for review

**Comment by grishchenkoa** [comment]
@hankovich 
```
Result<IEnumerable<TNew>> Combine<T, TNew>(this IEnumerable<Result<T>> results, Func<T, TNew> mapper)
```
looks like combination of two functions: `.Combine().Map(mapper)`

grishchenkoa:
> Overloads without errorMessagesSeparator should be added as well (You should use ", " as a default separator)
for this implementation i saw special task [https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/66](66)

Issue #70, opened 2018-10-16
Result Try method


kitsu:
I'm using your Result object with a third-party library that communicates using exceptions. I'd like to _lift_ the exceptions into a failed Result. My current code just returns a different Result from each branch of the try-catch.

I'm thinking something like:
``` CSharp
Result.Try(() => library.ThingThatMightFail(),
           (ex) => "Some error message, possibly branching on exception type...");

// Where
Try<Result<T>>(Func<T> attempt, Func<Exception,string> failure)
```

Handling specific exceptions can be done with `ex is SomeException`, and truly unexpected exceptions can be rethrown.

This would just be a convenience method, but often the reason I'm using a Result is to hide library exceptions behind more user friendly error messages.

freakingawesome:
I cobbled together a quick implementation of what this could look like.

```c#
// static method on Result
public static Result<T> Try<T>(Func<T> f, Func<Exception, Maybe<string>> handler)
{
    try
    {
        return Result.Ok(f());
    }
    catch (Exception x)
    {
        var handled = handler(x);
        if (handled.HasValue)
        {
            return Result.Fail<T>(handled.Value);
        }

        throw;
    }
}

// an extension for chaining
public static Result<T> OnSuccessTry<T>(this Result<T> self, Func<T> f, Func<Exception, Maybe<string>> handler) =>
    self.IsFailure ? self : Try(f, handler);
```

It's important to rethrow via `throw;` rather than `throw x;` because the `throw;` variation retains the original call stack, whereas `throw x;` creates a new call stack.

There could be lots of variations on this theme - async, handlers that return `Result`, etc.

ahryshchanka:
@vkhorikov is it make sense to implement this extension methods?

ahryshchanka:
@vkhorikov looks like this issue resolved and can be closed

PR #92, opened 2019-01-20
Add Result.Try method (#70)

ahryshchanka:


ahryshchanka:
PR for #70 
This PR is part of changes, other commits will be

hankovich:
I'm not very happy that I need to specify type args for `Try` method.
I don't want do `Result<int, int>.Try(() => 42, ex => 42)`, I want to do `Result.Try(() => 42, ex => 42)` and automatically get `Result<int, int>` as a return type.

vkhorikov:
Merged, thanks. Will be available in 1.16.0

Issue #93, opened 2019-01-24
ValueObject Equals throws Exception


linkdotnet:
What is the reason behind throwing an exception if the types of two ValueObjects don't match?
As stated by [MSDN about Object.Equals](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netframework-4.7.2): 

> Implementations of Equals must not throw exceptions; they should always return a value. For example, if obj is null, the Equals method should return false instead of throwing an ArgumentNullException.

I would propse just to return false. If you want I can create a PR.

vkhorikov:
Yes, I'm familiar with this implementation, and `ValueObject` had this implementation initially. I'm going back-and-forth on this issue, actually. The idea behind throwing an exception was that Value Objects are a special case. If you compare 2 value objects of different types, it always means there's a bug in the code, hence the code should fail fast.

Recently, I found a counter-example to this guideline, though, so need to revert this logic back to MS's canonical implementation, indeed.

Please, go ahead with the PR. Thanks.

linkdotnet:
As stated in the PR. There are cases where comparing different `ValueObject` of different types can be the case. Newtonsoft.Json uses `Equals` to determine circular references.
Yes it is arguable if serializing ValueObjects is not an anti-pattern itself and furthermore if using `Object.Equals` is the right way for determining circular references.

vkhorikov:
Published in 1.16.1, closing the issue

PR #95, opened 2019-01-24
ValueObject.Equals returns false in case of different types

linkdotnet:
`ValueObject.Equals(object other)` returns now false on different types instead of throwing ArgumentNullException.
The test was adopted as well.

The main benefit is we are closer to the expectation what `ValueObject.Equals` does.
See: [msdn](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netframework-4.7.2)
Furthermore now you can use JsonSerializer without risking to get an exception.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ValueObjectTests/BasicTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ValueObjectTests/BasicTests.cs
@@ -38,13 +38,12 @@ namespace CSharpFunctionalExtensions.Tests.ValueObjectTests
         }
 
         [Fact]
-        public void Comparing_value_objects_of_different_types_throws_an_exception()
+        public void Comparing_value_objects_of_different_types_returns_false()
         {
             var vo1 = new VO1("1");
             var vo2 = new VO2("2");
 
-            Action action = () => vo1.Equals(vo2);
-            action.ShouldThrow<ArgumentException>();
+            vo1.Equals(vo2).Should().BeFalse();
         }
 
         public class VO1 : ValueObject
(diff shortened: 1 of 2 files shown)


space-alien:
Vladimir, your [previous comment on this point](http://disq.us/p/1t7g1nu) caught my attention: "If you compare 2 value objects of different types, it always means that there's a bug in the code."

In light of this pull request, is it fair to say that you feel less strongly about this now? (Setting aside JsonSerializer whose use of Equals is discussed in [this SO answer](https://stackoverflow.com/a/51670641/).)

vkhorikov:
@space-alien that's correct. I actually found a counter-example and wanted to modify the VO's implementation back myself but @linkdotnet beat my on that.

Here's a bit more of this discussion: https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/93

The counter-example: https://enterprisecraftsmanship.com/2018/12/24/hierarchy-value-objects/

linkdotnet:
@vkhorikov Can you explain to me why (with regard to the linked article) this is a counter-example for using true/false pattern in the Equals-method? As I agree with your article, changing the (expected) behavior of Equals() depending on a special case seems like not the right way, or? For this particular case there are many possible solutions (like you stated, that you nest the VO in the Entity itself which can be easily achieved by NHibernate through component-mapping).
Or did I get something completely wrong?

PS: I like your open discussions. Keep this up!

vkhorikov:
@linkdotnet the linked article is a counter-example for throwing exceptions in `Equals`, not using true/false. Initially, I thought that inheritance of value objects was impossible/undesirable. Hence, the idea was to throw an exception when you compare 2 value objects of different types because the only valid case for that (aside from serialization which I didn't consider at the time) would be when you've got polymorphism and try to compare value objects within the same class hierarchy. Since this option was not on the table (due to the undesirability of value object inheritance), it was fine to throw -- it took value objects closer to how they are treated in F#.  In F#, you won't even be able to compile a code that compares two values of different types.

Then I discovered an example where the value object inheritance is a fine use case (the case with the Document base value object and Passport/IdentityCard sub-value objects from the article), hence I changed my mind on throwing exceptions in Equals.

PR #103, opened 2019-02-15
Custom ResultExceptions to replace InvalidOperationException.

### LINKED ISSUE #101: Throw a proprietary exception when accessing Value in failed state (or Error in success state)
Just an idea, what do you think?

I'm using a custom ExceptionFilterAttribute to catch and package various application errors into an appropriate HTTP error response for consumption by my web app.

I've found myself repeatedly checking my Results for success, and then throwing an exception on failure to pass the result error message to the filter.

However, Result already does this - attempting to access Value for a failed result throws an InvalidOperationException.

If, instead, Result threw a proprietary exception type (say, ResultFailureException) then I could skip the failure check and allow the filter to handle it with a 400 error instead of a 500.

If this sounds like a helpful change, I could submit a PR?

space-alien:
Fixes #101 

(with Error values, this time)

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/FailedResultTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/FailedResultTests.cs
@@ -34,7 +34,17 @@ namespace CSharpFunctionalExtensions.Tests.ResultTests
 
             Action action = () => { MyClass myClass = result.Value; };
 
-            action.ShouldThrow<InvalidOperationException>();
+            action.ShouldThrow<ResultFailureException>();
+        }
+
+        [Fact]
+        public void Cannot_access_Value_property_with_a_generic_error()
+        {
+            Result<MyClass, MyErrorClass> result = Result.Fail<MyClass, MyErrorClass>(new MyErrorClass());
+
+            Action action = () => { MyClass myClass = result.Value; };
+
+            action.ShouldThrow<ResultFailureException<MyErrorClass>>();
         }
 
         [Fact]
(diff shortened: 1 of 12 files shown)


space-alien:
Ah, drat, I've just noticed that this pull request #103 has accidentally included two extra commits from my fork which I must have pushed after the fact, not realising that it would be reflected on the PR. I don't think it's the end of the world though:

[The first commit](https://github.com/vkhorikov/CSharpFunctionalExtensions/pull/103/commits/7c344653ae720728622a570a19dec5d0e3fd9e26) removes the ConfigureAwait optional parameter as per #104 but doesn't add the global configurable property - I'll add that shortly with another PR.

[The other commit](https://github.com/vkhorikov/CSharpFunctionalExtensions/pull/103/commits/fa10e98a3d2662c3fe5cd006158d0657afbe27d5) changes the implementation of Map() to act as a simple pass-through for OnSuccess(). My impression from reading the code was that Map merely replicated OnSuccess, but I was still in the process of validating it before I planned to submit another issue+PR. Also I found one signature for Map that seemed to miss the TError in the return type, so I fixed that.

Please revert the merge if necessary. Otherwise I will prepare another PR shortly, as mentioned.

Finally, it's worth noting that removing the optional ConfigureAwait param is a potentially breaking change, and that I opted for `false` as a new default rather than the current `true`. The reason it was `true` is in #53, however, the scenario mentioned [may no longer be relevant since .NET 4.6](https://stackoverflow.com/questions/45530837/why-is-the-culture-kept-after-configureawaitfalse). Anyway, I was deliberating this for a while, but since the standard advice for libraries seems to be `false`, and we're already breaking (but with the configurable property as a workaround), it seemed reasonable.

vkhorikov:
This is fine, no need to revert the commit. Regarding the ConfigureAwait param - it is technically a breaking change but it can easily be fixed by using the new static property.

PR #116, opened 2019-03-28
Serializable attribute on Result classes

jcsonder:
Add serializable attribute and reqiured constructor for deserialization to Result classes.


jcsonder:
Created a new PR with fixed user/only one.

PR #117, opened 2019-03-29
SerializableResultClasses

jcsonder:
Add serializable attribute and required constructor for deserialization to Result classes.


jcsonder:
Needs some improvements according: https://docs.microsoft.com/en-us/dotnet/standard/serialization/serialization-guidelines

PR #118, opened 2019-03-31
Serializable/Deserializable result classes

jcsonder:
Add serializable attribute and reqiured constructor for deserialization to Result classes.


vkhorikov:
Merged, thanks. Published in 1.18.0.

jcsonder:
Thank you! 👍

PR #123, opened 2019-04-22
add support for maybe linq-like select chaining

pilotMike:
allows 
```
Maybe<int> result = from a in maybe1 from b in maybe2 select a + b;
```


pilotMike:
allows 
```
Maybe<int> result = from a in maybe1 from b in maybe2 select a + b;
```

vkhorikov:
Merged, thanks.

SuperJMN:
Is this available for Result<T> and Result<T,E>

E.g.

```csharp
var query = from r in someResult
  from s in someOtheResult
  select new { r, s };
```

hankovich:
@SuperJMN It's available for `Result<T>` and `Result<T, E>`, but `Result` and `UnitResult<E>` are not (obviously) supported.

PR #127, opened 2019-06-23
Remove generic constraints requiring errors to be reference types.

ergwun:
Null errors for results representing failures will still throw an
exception, but errors that are structs can use the default value
without causing an exception to be thrown.


vkhorikov:
Merged, thanks!

PR #128, opened 2019-07-16
Match extension method

dpraimeyuu:
This PR introduces possibility to use values kept in either `Maybe` or `Result` in a safe and functional way. It forces handling all possible paths by bringing `match`-like behavior from F#.

References: [1](https://blog.ploeh.dk/2018/06/04/church-encoded-maybe/), [2](https://blog.ploeh.dk/2018/06/11/church-encoded-either/)

Btw. @vkhorikov I love your videos, blog and this package, so thank you for all great work!


vkhorikov:
Merged, thanks :)

vkhorikov:
Fun fact: I was considering both implementations (Either/Match in the F# style) and .OnSuccess/.OnFailure this lib's style but decided that F# style is too complicated. Now I think that it's actually cleaner the F# way :)

Issue #126, opened 2019-06-17
Why is the generic error type constrained to be a reference type?


ergwun:
The following code is not permitted, since `Result.Fail<T, E>(E error)` constrains the generic type parameter E to be a reference type:

var myErrorCode = 1;
var result = new Result.Fail<string, int>(myErrorCode);

Why do the various static factory methods for creating an instance of `Result<T, E>` have this constraint (especially given that the `Result<T, E>` struct itself does not)?

vkhorikov:
It's to make sure the preconditions like `if (error == null)` in `ResultCommonLogic` work. I'm fine with changing this as long as those preconditions will continue working (i.e a client shouldn't be able to create a failed result without an error). Feel free to submit a PR.

ergwun:
Thanks. PR submitted. I included a bunch of unit tests to ensure that the checks for null errors still work.

vkhorikov:
A new version should be listed on nuget within the next couple hours: 1.18.2
Thanks

BillyFromAHill:
Looks like there is related issue now. Error has value ```default (E)``` in case of using ```Ok<T, E>(T value)```. But after that in ResultCommonLogic we have ```else if ((object) error != null)``` so we can't create Result with enum error type. I think, this condition should compare error with default (E).

vkhorikov:
Could you post a test that reproduces the issue?

BillyFromAHill:
Yes. Just ```Result.Ok<int, int>(0);``` throws ArgumentException from here https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/0e70ff6718e2dac8e1cfa71b93aec6d63d4dfb81/CSharpFunctionalExtensions/Result.cs#L40-L41

vkhorikov:
Apology for the late reply. Please do create a PR. Thanks much.

PR #133, opened 2019-08-01
Fix ArgumentException with struct error type.

BillyFromAHill:
Check condition in case of Ok Result made appropriate for value types.


vkhorikov:
Thanks! I'll push a new version this weekend.

space-alien:
See #126

PR #145, opened 2019-09-03
Combine improvements

space-alien:
* All Combine signatures now have an optional error message separator param.
* DRYed up and improved the code
* Added some docs


CSharpFunctionalExtensions/Result/Extensions/CombineAsyncLeft.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace CSharpFunctionalExtensions
```
golavr: you can remove the platform divergence in functions using: 

```
#if NET40
using Task = System.Threading.Tasks.TaskEx;
#else
using Task = System.Threading.Tasks.Task;
#endif
```
image

golavr:
to make combine more fluent in some cases
meaning not having to create collection of results or tasks results to be able using combine

you can add:

image



image

...need to add all variations

vkhorikov:
@space-alien looks great, merge as you see fit.

space-alien:
@golavr I like the look of the Task<Result> extension methods too, I'd like to loop back on this when I have a bit more time.

PR #156, opened 2019-09-17
Add missing Result.Try for async action

CommCody:
We are missing this `Result.Try` overload in our use of Result

The reason is that if we don't have this overload, the overload resolution fallbacks to the  `... Result<T> Try<T>(Func<T> func, ...` overload and this results in a `Result<Task>` return type.


vkhorikov:
Merged, thanks!

PR #162, opened 2019-10-12
Add summary tag to Bind method

bopazyn:
The methods were decorated with text

`Selects result from the return value of a given function. If the calling Result is a failure, a new failure result is returned instead.`

If you have better description, what methods do, i will change it


vkhorikov:
Merged, thanks.

PR #163, opened 2019-10-12
Add proposition for dealing with tasks

bopazyn:
How community will react for this method?. Are they usuable? I would like to start the discussion.

### DIFF
--- a/CSharpFunctionalExtensions/Task/Bind.cs
+++ b/CSharpFunctionalExtensions/Task/Bind.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace CSharpFunctionalExtensions
+{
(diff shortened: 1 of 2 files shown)


vkhorikov:
It's better to create an issue for the conversion, PRs are for code change proposals. You can link this PR to an issue for reference.

Regarding `TaskExtensions` themselves, I don't think there's a lot of value in them. The C# compiler automatically wraps async methods' return value into a `Task`, and so unlike `List`, `Maybe`, or `Result`, you can handle the `Task` monad using native C# language syntax -- which is simpler and more familiar for most people.

bopazyn:
At first glance this methods made sens but after trying using them i realised that my code is hard to read and understand. This idea will definitely need more time in my head.

I am closing this PR and will try to find cases where the methods would be usefull.

PR #165, opened 2019-10-23
Rename Maybe's method from Select to SelectMany

bopazyn:
Signature of this method suggests it should be named `SelectMany`.


vkhorikov:
The old `Select` overloads need to be obsoleted so that people using the older versions of the lib have time to upgrade. Also, `Select` and `SelectMany` need to call `Map` and `Bind` internally.

bopazyn:
The PR was updated. I used `Bind` and `Map` methdods from PR !167 to remove duplicate code.

PR #168, opened 2019-10-23
Add SelectMany method for Result

bopazyn:


vkhorikov:
Could you delegate the work in `SelectMany` to `Bind`? No need to reimplement ``Bind``'s functionality.

bopazyn:
PR was updated. What was interesting for me is the fact, that i could replace logic in `SelectMany` with just one `Bind` and one `Map`.

vkhorikov:
Yep, different aspects of programming sometimes converge in an unexpected and fascinating way.

Merged, thanks!

PR #169, opened 2019-10-24
Implicit Maybe bool conversion

Schwefelhexa:
Allows for `Maybe`s to be converted to `bool`s implicitly.
_Should_ include all tests required.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
@@ -13,6 +13,7 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
 
             maybe.HasValue.Should().BeFalse();
             maybe.HasNoValue.Should().BeTrue();
+            ((bool) maybe).Should().BeFalse();
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


vkhorikov:
What is the use case for this conversion?

Schwefelhexa:
It's just meant as a shorthand instead of writing `maybe.HasValue` to prevent repetitive code.

vkhorikov:
Do you have some examples of where this is useful? I have a hard time imagining this would be more readable than `maybe.HasValue` but maybe I'm missing something.

Schwefelhexa:
I honestly just felt like it would prevent you from writing unnecessary code.

For example
```C#
var result = SomeMethod():
if(result) {
    // Success!
} else {
    // Failure :(
}
```
seems still very readable to me. I'd even argue it's more readable since it's "compressing" the code to into less space and I personally think it's still obvious what the code is meant to do.

vkhorikov:
A better approach would be to introduce a `Deconstruct` method like the one for `Result`: https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/9f9577d9a2b2bb15c5f2e684c51e7be3c135c996/CSharpFunctionalExtensions/Result/Extensions/Deconstruct.cs (although, just one extension method would be enough). The code will be somewhat similar:

```
(var hasValue, _) = SomeMethod();
```

but with an explicit 2nd value.

With implicit conversions to a single primitive type, it's easy to lose sight of the second Maybe's attribute (`Value`).

Issue #184, opened 2020-03-01
Overload for TapIf with Func<T, bool> condition


pedromtcosta:
Currently the `TapIf` extensions only accept a `bool` parameter for the condition. I think it would be convenient to also have overloads that accept a `Func<T, bool>` and `Func<T, E, bool>` for the condition. Should these overloads be added, or are there any reasons for why they're not there in the first place?

vkhorikov:
Can you give an example of one such overload implemented?

pedromtcosta:
Example of the overload implemented, with and without `async`:

```csharp
public static Result<T> TapIf<T>(this Result<T> result, Func<T, bool> condition, Action action)
{
    if (result.IsSuccess && condition(result.Value))
        return result.Tap(action);
    else
        return result;
}

public static async Task<Result<T>> TapIf<T>(this Task<Result<T>> resultTask, Func<T, bool> condition, Func<Task> func)
{
    Result<T> result = await resultTask.ConfigureAwait(Result.DefaultConfigureAwait);

    if (result.IsSuccess && condition(result.Value))
        await func().ConfigureAwait(Result.DefaultConfigureAwait);

    return result;
}
```
And here are the usage examples:

```csharp
Result.Success(new Customer())
    .TapIf(c => string.IsNullOrEmpty(c.Email), Do);

await Task.FromResult(Result.Success(new Customer()))
    .TapIf(c => string.IsNullOrEmpty(c.Email), DoAsync);
```

FrancoisM:
IMO, i'd make more sense to define a Where extension and then do Where(...).Tap(...).

pedromtcosta:
I might be missing something, but wouldn't the `Where()` method have the same behavior as `Ensure()` in this case?

vkhorikov:
Ah, I see the use case now. Looks good. No particular reason for not implementing it, feel free to submit a PR.

A `Where` extension wouldn't work here because it would loose the context of what needs to be done with the incoming result.

pedromtcosta:
I will start working on that and should submit the PR on the weekend.

I am considering in which of these 2 is a better way to implement these methods:
```csharp
if (result.IsSuccess && condition(result.Value))
    action();

return result;
```

```csharp
if (result.IsSuccess && condition(result.Value))
    return result.Tap(action);
else
    return result;
```

[...]

PR #185, opened 2020-03-08
TapIf with Func<T, bool> overloads

pedromtcosta:
Adds TapIf overloads with Func<T, bool> condition.

Discussion on: #184


vkhorikov:
Looks good. I'll take a closer look in the evening and will merge then.

vkhorikov:
Merged, thanks!

Issue #151, opened 2019-09-06
Proposal to make Result a class instead of a struct


space-alien:
`Result` is currently a struct. (Actually, three structs, because of `Result<T>` and `Result<T, E>` as well.)

Unfortunately, this creates rather a lot of complications.

1. Implicit conversion operators are required to convert between the different structs, since there is no inheritance. In particular, the implicit conversions from `Result<T, E>` to `Result<T>` / `Result` actually call `ToString()` on the `E` error value, which seems rather presumptuous. Some Map() overloads rely on this conversion to get from one Result type to another, so custom errors could conceivably lose fidelity in the process.

2. The logic for creating Results is encapsulated in a special class `ResultCommonLogic`, an instance of which is then stored on every `Result`. This adds complexity and overhead. In spite of this, there is inevitably still plenty of code duplication.

3. Issue #125 highlights the requirement for a non-value `Result` with a custom error type, something that would add yet another struct to the mix, whereas a base class might be a more fitting solution.

4. There are challenges ahead with managing and improving extension overloads (see #146), some of which might be alleviated with some judicious use of class inheritance. (I am still investigating this).

I propose switching `Result` back to a class and leveraging inheritance to solve these problems. Based on my experiments so far, I think we might achieve a significant reduction in complexity, and improve maintainability.

Here are two possible approaches. Note how in both cases the base class is a new non-value result with generic `E` error type, solving #125):

**(1)** Maintain current types

```csharp
class ResultBase<E> {}
class Result : ResultBase<string> {}
class Result<T, E> : ResultBase<E> {}
class Result<T> : Result<T, string> {}
```

**(2)** This may be cleaner and more intuitive, but will break more
```csharp
class Result<E> {}
class Result : Result<string> {}
class ValueResult<T, E> : Result<E> {}
class ValueResult<T> : ValueResult<T, string> {}
```

I have started playing around with an implementation of the first approach and I am very encouraged, with complexity falling away as I progress. I still want to experiment with the second syntax though.

#99 is another (minor) issue that would be solved by this change. There are probably other benefits I haven't identified yet.

vkhorikov:
I agree, switching `Result` to a `struct` has led to a lot of issues. The only problem I see with switching it back is the loss of non-nullability, but considering the complexity reduction, this trade-off is worth it.

I like (1) better due to backward compatibility. Also, I don't quite like the word `Value` in `ValueResult`, looks bulky.

[...]

PR #186, opened 2020-03-09
Quality improvements

space-alien:
Along with some package updates, this PR addresses the first point of #151, namely the risky implicit conversion operators from `Result<T, E>` to the other types.

Whether or not we make a switch to classes for `Result` or find a way to use interfaces to simplify things (I'm still investigating these questions), these implicit conversions seem like a mistake. Implicitly converting `E` to `string` via `ToString()` definitely feels wrong.

Removing these operators even shook out a few minor bugs/mistakes in the library itself, perhaps hinting at how easily they can be misused inadvertently.

Technically this is a breaking change, although I would think that any breakages are likely to be the good kind... but keen to hear your thoughts.

### DIFF
--- a/CSharpFunctionalExtensions.Examples/ResultExtensions/ExampleWithOnFailureMethod.cs
+++ b/CSharpFunctionalExtensions.Examples/ResultExtensions/ExampleWithOnFailureMethod.cs
@@ -28,7 +28,7 @@ namespace CSharpFunctionalExtensions.Examples.ResultExtensions
         {
             public void AddBalance(decimal moneyAmount)
             {
-                
+
             }
         }
 
@@ -41,7 +41,7 @@ namespace CSharpFunctionalExtensions.Examples.ResultExtensions
 
             public void RollbackLastTransaction()
             {
-                
+
             }
 
             public Task RollbackLastTransactionAsync()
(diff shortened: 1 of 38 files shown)


vkhorikov:
I don't use `Result<T, E>` very much, so can't say from personal experience.

What about the implicit conversion from `Result<T>` to `Result`? Could you formulate what makes the conversion from `Result<T, E>` to `Result<T>` different?

space-alien:
I *think* the implicit conversion from `Result<T>` to `Result` shouldn't be dangerous. Although the `Value` is discarded, consumers of `Result` shouldn't miss it since `Result` exposes no value.  If we were using classes rather than structs, `Result<T>` would be derived from `Result` so this conversion mimics implicit upcasting.

Whereas conversions from `Result<T, E>` to `Result<T>` or `Result` _are_ dangerous, because now we are changing a custom `E` error class into a `string` without warning, and presuming that the user has implemented a suitable ToString() method. I personally haven't used custom error classes, but I imagine that if I were, this is not something I would want happening.

space-alien:
Further to the above, I've added some documentation for the Result methods, along with some minor code formatting commits.

vkhorikov:
Your reasoning makes sense. Feel free to merge this to master.

PR #187, opened 2020-03-11
Async code improvements, C#8, Bind() unit tests

space-alien:
Please cast your eye over these ideas; I expect there will be similar opportunities on other extension methods.

I don't see why `BindAsyncRight` needs to await.

There is now much better alignment/reuse between `BindAsyncBoth`/`BindAsyncLeft`, and between `BindAsyncRight`/`Bind`.


vkhorikov:
Looks good, nice simplifications

Issue #181, opened 2020-02-20
Tap seems limited since you cannot halt execution on failure


benprime:
Tap is very useful in that you can use it to return the passed in result and do multiple operations on a single result object. For example.

```
return getMyResult()
    .Tap(x => DoWork1(x))
    .Tap(x => DoWork2(x))
    .Tap(x => DoWork3(x));
```

However, sometimes you would like to return the result and still depend on whether the DoWork() method failed. If you could assume that `Func`s passed to DoWork all returned the same result type as the "parent" result, that would be useful.

I find myself writing extensions methods like the following:

```
public static async Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Task<Result<T>>> func)
{
	var result = await resultTask.ConfigureAwait(Result.DefaultConfigureAwait);
	if (!result.IsSuccess) return result;
	var tapResult = await func(result.Value).ConfigureAwait(Result.DefaultConfigureAwait);
	return tapResult.IsFailure
		? tapResult
		: result;
}
```

In this case, the flow would be identical, except that execution would stop when a tap `Func` returns a failed result.

What is the reasoning behind having the result of a Tap operation be ignored in the railway flow?

vkhorikov:
You should be able to use `Bind` for this. `Tap` is only for when you want to ignore the result entirely.

The reason for the separation between `Bind` and `Tap` (previously, they all were overloads of `OnSuccess`) is to separate `OnSuccess` due to it becoming too bloated.

benprime:
I have not come up against many scenarios where I'd want to ignore the result completely, but I have come up against several scenarios where I would like to run an operation, check the result as normal, but return the original result rather than the new result.

Unless I'm missing some context, using bind for this would amount to this code, correct?

```
return getMyResult()
    .Bind(x => {
        var result = DoWork1(x);
        if (result.IsSuccess) return Result.Success(x);
        return result;
    })
    .Bind(x => {
        var result = DoWork1(x);
        if (result.IsSuccess) return Result.Success(x);
        return result;
    })
    .Bind(x =>
    {
        var result = DoWork1(x);
        if (result.IsSuccess) return Result.Success(x);
        return result;
    });
```

This works, but the developers that I work with (and myself) usually end up writing extension methods like `TapAndCheck` that feel like they should be core to the library.

space-alien:
I do it this way:

```csharp
return GetCustomer()
[...]

PR #188, opened 2020-03-12
Tap()/TapIf() - Major overhaul including examples.

space-alien:
Solves #174 and #179.

A major overhaul of Tap(). **42*** unhelpful methods and their unit tests removed with no obvious sacrifices.

(* I rather like the symbolism, don't you?!)

PR #189, opened 2020-03-12
Adding Result.TapAndCheck extension method

benprime:
Adding TapAndCheck extension method as discussed in #181


space-alien:
Thanks for this Ben, it looks promising.

1. Regarding the `Bind()` overloads you've added; we removed those last year because the starting result value is silently discarded, which may be risky. Instead, you can achieve the same outcome with: `result.Bind(_ => someFunc())`; and it now is clear that you intended to discard the starting value.

2. Same applies to that `Map()` overload.

3. `TapAndCheckTests` should have their own base `TapAndCheckTestsBase`. I suggest you hang on until #188 is merged to avoid going over the tests twice.

4. Presumably, this:
```
if (result.IsFailure)
    return Result.Failure<T>(result.Error);  // Or whatever
```

Should always simplify to this, because the return type never changes?
```
if (result.IsFailure)
    return result;
```

5. Sometimes you are returning like this: `return func().Map(() => result.Value);` sometimes like this: `return funcResult.IsSuccess ? result : Result.Failure(result.Error);` The latter statement is going to break, because `result` will be a success in that case, and accessing `Error` will throw. Actually, in case of `Result` you could just `return func()`!

6. In terms of signatures, I think these are all the current possible permutations:
```
From             To

Result       | Func<Result>
Result       | Func<Result<K>>
Result       | Func<Result<K, E>>
Result<T>    | Func<Result>
Result<T>    | Func<Result<K>>
Result<T>    | Func<Result<K, E>>
Result<T>    | Func<T, Result>
Result<T>    | Func<T, Result<K>>
Result<T>    | Func<T, Result<K, E>>
Result<T, E> | Func<Result>
Result<T, E> | Func<Result<K>>
Result<T, E> | Func<Result<K, E>>
Result<T, E> | Func<T, Result>
Result<T, E> | Func<T, Result<K>>
Result<T, E> | Func<T, Result<K, E>>
```

The signatures would need careful realignment with the above. For example, sometimes there is `T` in place of `K` - this would prevent the overload from resolving if the tapped method returns a different value type from the starting result.

That's as far as I've got for now, hope this is helpful.

@vkhorikov Upon seeing the sheer number of overloads here, I'm now starting to think that this does seem like a big addition when there is a simple usage syntax we could add to the examples: `result.Bind(c => UpdateCustomer(c).Map(() => c))`.

benprime:
@space-alien Thanks for the feedback. I'll make the changes as suggested and hold off on testing changes until #188 is merged.

space-alien:
Thanks Ben. Meanwhile, I've marked some of the signatures above with asterisks. I'm wondering whether we need all of these ones - not sure how useful they are. Will need to think about this some more tomorrow.

benprime:
@space-alien All good points, and thanks again for your feedback.

Some questions/comments:

> 1. Regarding the `Bind()` overloads you've added; we removed those last year because the starting result value is silently discarded, which may be risky. Instead, you can achieve the same outcome with: `result.Bind(_ => someFunc())`; and it now is clear that you intended to discard the starting value.
> 2. Same applies to that `Map()` overload.

I wasn't sure about adding these, honestly. I followed your suggestion and removed them and have now got it working the way you recommended.

> 1. Sometimes you are returning like this: `return func().Map(() => result.Value);` sometimes like this: `return funcResult.IsSuccess ? result : Result.Failure(result.Error);` The latter statement is going to break, because `result` will be a success in that case, and accessing `Error` will throw. Actually, in case of `Result` you could just `return func()`!

This was actually a typo on my part. I meant 
`return funcResult.IsSuccess ? result : Result.Failure(result.Error);`
to be
`return funcResult.IsSuccess ? result : Result.Failure(funcResult.Error);`
I don't think we want to return `func()`, since this is a "Tap" and none of these overloads will return the result of the `func`. Does that make sense?

space-alien:
Although it would be okay to return `func()` where we've got a `Result` (Results are structs, so it would be an identical value), what is more important is finding common (or identical) syntax to use for all the overloads, to improve maintainability.

benprime:
> Although it would be okay to return `func()` where we've got a `Result` (Results are structs, so it would be an identical value), what is more important is finding common (or identical) syntax to use for all the overloads, to improve maintainability.

I see. My concern was more that you would break the behavior of the Tap by returning `func()`. In a Tap, you provide a func, but don't want to return the result, so in returning `func()`, I'd just essentually be doing a `Bind()` wouldn't I?

On the question of which overloads are appropriate, my thinking was since Tap will execute, and the result would be thrown away, that it makes sense to allow all func signatures, since mismatching didn't seem to be a problem. However, looking at your list of signatures and re-reviewing the code, I think that these signatures should be removed:

```
Result       | Func<Result<K, E>> *
Result<T>    | Func<Result<K, E>> *
Result<T>    | Func<T, Result<K, E>> *
```
In any signature where we have E on the func and not in the left column, we are not going to be able to return a failed result since we would be casting from E to string.

I may be way off base here, please let me know if any of this is wrong.

benprime:
@space-alien I have refactored the methods to match the signature list:
```
Result       | Func<Result>
Result       | Func<Result<K>>
Result<T>    | Func<Result>
Result<T>    | Func<Result<K>>
Result<T>    | Func<T, Result>
Result<T>    | Func<T, Result<K>>
Result<T, E> | Func<Result<K, E>>
Result<T, E> | Func<T, Result<K, E>>
```
I re-ordered the methods to match the list so they will be easy to review.

Testing may need some review, and I will do that once #188 is merged.

space-alien:
> > Although it would be okay to return `func()` where we've got a `Result` (Results are structs, so it would be an identical value), what is more important is finding common (or identical) syntax to use for all the overloads, to improve maintainability.
> 
> I see. My concern was more that you would break the behavior of the Tap by returning `func()`. In a Tap, you provide a func, but don't want to return the result, so in returning `func()`, I'd just essentually be doing a `Bind()` wouldn't I?

TapAndCheck is effectively a `Bind()` followed by a `Map()` back to the starting value for the success case.

In the cases where we return `Result` (no T value), there is no value to Map, so the code is indeed identical to `Bind()`.

So where you have this:
```csharp
public static Result TapAndCheck(this Result result, Func<Result> func)
{
    if (result.IsFailure)
        return result;

    var funcResult = func();
    return funcResult.IsSuccess
        ? result
        : Result.Failure(funcResult.Error);
}
```
`funcResult` contains a `Result` (no T value). Because the Result types are structs (i.e. value types), if `funcResult` is a success then it already contains the exact same **value** as `result` (there is no reference to an object to preserve). If `funcResult` is a failure, then again it is already the value we want to return: a failed `Result` with the error message we wanted to capture. Calling `Result.Failure(funcResult.Error)` does nothing except return the same `funcResult`. So these four lines should be replaced with `return func();`

As for `Result TapAndCheck<K>(this Result result, Func<Result<K>> func)`, in this case the compiler calls the implicit cast operator from `Result<T>` to `Result`.

> On the question of which overloads are appropriate, my thinking was since Tap will execute, and the result would be thrown away, that it makes sense to allow all func signatures, since mismatching didn't seem to be a problem. However, looking at your list of signatures and re-reviewing the code, I think that these signatures should be removed:
> 
> ```
> Result       | Func<Result<K, E>> *
> Result<T>    | Func<Result<K, E>> *
> Result<T>    | Func<T, Result<K, E>> *
> Result<T, E> | Func<Result> *
> Result<T, E> | Func<Result<K>> *
> Result<T, E> | Func<T, Result> *
> Result<T, E> | Func<T, Result<K>> *
> ```
> 
> In any signature where we have `E` on the left and not on the right (or vice versa) , we are not going to be able to return a failed result since we would be casting between `E` and `string`.

Exactly!

benprime:
> So these four lines should be replaced with `return func();`
Ah! Sorry, I follow you now. That does make sense. My only concern would be developers trying to make sense of the library by viewing the source code. I often look a the simplest overloads when I'm looking to see what a method does. In this case, they may draw the wrong conclusions until they looked at more complicated overloads. Perhaps that is a minor concern.

space-alien:
Quite right. But I would say it's not a concern in this case. Furthermore, `return func()` is actually the same as the other overloads, just without the tailing `.Map` which is needed for the `Result<T>`s. So this is much better for code consistency between overloads, which is more important for maintainability and also helps a new developer when learning the library.

vkhorikov:
Apology for the late replies. I've merged #188 

@space-alien , moving forward, don't hesitate to merge PRs yourself unless you want me to take a close look or there are some questions that need an additional eye on them -- this would remove myself as a bottleneck. I'll be reviewing such PRs asynchronously.

vkhorikov:
Should there be `TapAndCheck` for `Result` (no T) at all? Since there's no need to preserve the original `Result` instance, one could call `Bind` instead.

I suggest to either remove these overloads or call `Bind` inside of it. Don't have a strong opinion here.

space-alien:
1. Good point! Yes, I think we should remove them, to help reduce duplication and the potential confusion there.

2. I'm starting to think we should add these new methods as `Tap()` overloads. My suspicion is that, more often than not, I'd want to catch a failure when I tap a `Result`-returning method. As it stands, the user will have to remember to use this new `TapAndCheck` method when tapping a method that returns a Result. This seems to run counter-intuitively to the way things normally work in the library.

   If we do *really* want to ignore an inner failed Result, the syntax is easy enough and also makes the intent more explicit: `result.Map(v => { TapCarelessly(); return v; })`.

3. Finally, there crept in a few signatures that do not supply the starting result value as a parameter, something we have steered away from elsewhere. This is my bad, since I missed this when I was thinking through all the possible permutations the other night. So I think these signatures should also be removed:

   ```csharp
   Result<T> TapAndCheck<T>(this Result<T> result, Func<Result> func)
   Result<T> TapAndCheck<T, K>(this Result<T> result, Func<Result<K>> func)
   Result<T, E> TapAndCheck<T, K, E>(this Result<T, E> result, Func<Result<K, E>> func)
   ```

    Instead, the recommended usage is a discard lambda param: `result.Tap(_ => TapMeBabyOneMoreTime())`

vkhorikov:
2. Good point, and I tend to agree. @benprime , thoughts? I don't use `Tap` much, so can't tell from my own experience.

3. I agree.

`TapMeBabyOneMoreTime` :smile:

benprime:
1. Agreed. These overloads don't really need to exist, I'll remove them.
2. This is a good idea. I agree that if `Tap()` is passed a `Func` that returns a `Result`, then it should behave as `TapAndCheck` currently does in this PR. I can refactor this PR to be `Tap` overloads if that is the consensus. However, this may be dangerous, as there may be code out in the wild that is currently passing in `Result`-returning `Funcs` to `Tap` and may be expecting the `Func` result to be ignored pass or fail.
3. That's a fair point, I will remove these overloads.

benprime:
@vkhorikov @space-alien I've pushed the changes we discussed. I have not re-added tests yet, but I wanted to confirm that these are the only appropriate overloads for `Tap()` to supply the "TapAndCheck-style" behavior.

space-alien:
I think we are missing these signatures:

```csharp
public static Result Tap<K>(this Result result, Func<Result<K>> func)
{
    if (result.IsFailure)
        return result;

    return func(); // implicit cast
}

public static Result<T> Tap<T>(this Result<T> result, Func<T, Result> func)
{
    if (result.IsFailure)
        return result;

    return func(result.Value).Map(() => result.Value);
}
```

Agreed - this is a potentially breaking change, but I think it's still the way to go.

space-alien:
Edited that to remove the first signature, which we already talked about.

CSharpFunctionalExtensions/Result/Extensions/TapAsyncBoth.cs
```diff
@@ -69,5 +69,35 @@ public static async Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func
 
             return result;
         }
+
+        /// <summary>
+        ///     Executes the given action if the calling result is a success. Returns the calling result on success.
+        /// </summary>
+        public static async Task<Result> Tap<T>(this Task<Result<T>> resultTask, Func<T, Task<Result>> func)
```
space-alien: We need to remove this one

CSharpFunctionalExtensions/Result/Extensions/TapAsyncLeft.cs
```diff
@@ -49,5 +49,24 @@ public static async Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Acti
...
         }
+
+
+        /// <summary>
+        ///     Executes the given action if the calling result is a success. Returns the calling result on success.
+        /// </summary>
+        public static async Task<Result<T>> Tap<T, K>(this Task<Result<T>> resultTask, Func<T, Result<K>> func)
+        {
+            Result<T> result = await resultTask.DefaultAwait();
+            return result.Bind(func).Map(_ => result.Value);
```
space-alien: Can we not just `return result.Tap(action);` like the others here?

CSharpFunctionalExtensions/Result/Extensions/TapAsyncRight.cs
```diff
@@ -59,5 +59,27 @@ public static async Task<Result<T>> Tap<T>(this Result<T> result, Func<T, Task>
...
+        ///     Executes the given action if the calling result is a success. Returns the calling result on success.
+        /// </summary>
+        public static async Task<Result<T>> Tap<T, K>(this Result<T> result, Func<T, Task<Result<K>>> func)
+        {
+            if (result.IsFailure)
+                return result;
+
+            return await func(result.Value).Map(_ => result.Value);
+        }
+
```
space-alien: I think this simplifies down to:

```csharp
public static async Task<Result<T>> Tap<T, K>(this Result<T> result, Func<T, Task<Result<K>>> func)
{
    return await result.Bind(func).Map(_ => result.Value).DefaultAwait();
}
```

CSharpFunctionalExtensions/Result/Extensions/TapAsyncBoth.cs
```diff
@@ -69,5 +69,35 @@ public static async Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func
...
+
+        /// <summary>
+        ///     Executes the given action if the calling result is a success. Returns the calling result on success.
+        /// </summary>
+        public static async Task<Result<T>> Tap<T, K>(this Task<Result<T>> resultTask, Func<T, Task<Result<K>>> func)
+        {
+            Result<T> result = await resultTask.DefaultAwait();
+            var funcResult = await result.Bind(func).DefaultAwait();
+            return funcResult.Map(_ => result.Value);
+        }
```
space-alien: See comment on `TapAsyncRight`. Couldn't we do this?

```csharp
public static async Task<Result<T>> Tap<T, K>(this Task<Result<T>> resultTask, Func<T, Task<Result<K>>> func)
{
    Result<T> result = await resultTask.DefaultAwait();
    return await result.Bind(func).Map(_ => result.Value).DefaultAwait();
}
```

CSharpFunctionalExtensions/Result/Extensions/TapAsyncBoth.cs
benprime: > **EDITED**
> I think we are missing this signature:
> ```cs
> public static Result<T> Tap<T>(this Result<T> result, Func<T, Result> func)
> {
>     if (result.IsFailure)
>         return result;
> 
>     return func(result.Value).Map(() => result.Value);
> }
> ```
> 
Shouldn't all the overloads have a `Func` that returns simply a `Result`:

Are we also missing these then?
```
Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Task<Result>> func)
Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Result> func)
Task<Result<T>> Tap<T>(this Result<T> result, Func<T, Task<Result>> func)
```

benprime:
@space-alien 
> **EDITED**
> I think we are missing this signature:
> ```cs
> public static Result<T> Tap<T>(this Result<T> result, Func<T, Result> func)
> {
>     if (result.IsFailure)
>         return result;
> 
>     return func(result.Value).Map(() => result.Value);
> }
> ```
> 
Shouldn't all the overloads have a `Func` that returns simply a `Result`:

Are we also missing these then?
```
Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Task<Result>> func)
Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Result> func)
Task<Result<T>> Tap<T>(this Result<T> result, Func<T, Task<Result>> func)
```

CSharpFunctionalExtensions/Result/Extensions/TapAsyncBoth.cs
benprime: Because it doesn't return `Task<Result<T>>` ? Okay, removing.

CSharpFunctionalExtensions/Result/Extensions/TapAsyncLeft.cs
benprime: I thought that `Tap` signatures with `Result`-returning `Func`s were the only signatures that would behave like `TapAndCheck`, where the `Action` signatures would continue to ignore the result altogether. Wouldn't changing this to `return result.Tap(action)` cause us to call the wrong overload?

space-alien: No, the `func` parameter is still a `Func` and this just passes it to the non-async `Tap()` overload.

space-alien: I should have said `return result.Tap(func)` of course - sorry, missed that typo.

space-alien:
> @space-alien
> 
> > **EDITED**
> > I think we are missing this signature:
> > ```cs
> > public static Result<T> Tap<T>(this Result<T> result, Func<T, Result> func)
> > {
> >     if (result.IsFailure)
> >         return result;
> > 
> >     return func(result.Value).Map(() => result.Value);
> > }
> > ```
> 
> Shouldn't all the overloads have a `Func` that returns simply a `Result`:
> 
> Are we also missing these then?
> 
> ```
> Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Task<Result>> func)
> Task<Result<T>> Tap<T>(this Task<Result<T>> resultTask, Func<T, Result> func)
> Task<Result<T>> Tap<T>(this Result<T> result, Func<T, Task<Result>> func)
> ```

Yes, that's correct.

CSharpFunctionalExtensions/Result/Extensions/TapAsyncLeft.cs
benprime: Got it. Thanks for the clarification.

CSharpFunctionalExtensions/Result/Extensions/Tap.cs
```diff
@@ -58,5 +58,27 @@ public static Result<T> Tap<T>(this Result<T> result, Action<T> action)
...
+        /// <summary>
+        ///     Executes the given action if the calling result is a success. Returns the calling result on success.
+        /// </summary>
+        public static Result<T> Tap<T, K>(this Result<T> result, Func<T, Result<K>> func)
+        {
+            if (result.IsFailure)
+                return result;
+
+            return func(result.Value).Map(_ => result.Value);
+        }
```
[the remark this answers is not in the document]
benprime: Updated.

CSharpFunctionalExtensions/Result/Extensions/TapAsyncBoth.cs
benprime: Updated.

space-alien:
Thanks Ben, looks like we're getting there!

I've created a new topic branch (tap_result_funcs) so we can finish this on there.

By the way, I think it may be best not to force push updates to pull requests, because then we can't go back to review the changes between reviews.

space-alien:
Great, I've pushed one tidy up commit to the `tap_result_funcs` branch. Please check it out and when you are ready with the unit tests, you can open a new PR against that branch.

PR #191, opened 2020-03-20
Adding unit tests for tap result funcs

benprime:
Here I have provided unit tests for the result-returning tap signatures. I attempted to follow the patterns you had in place. Please let me know if you would like any changes.

@space-alien  @vkhorikov Question: Do we also need to add the func signatures to `TapIf`?


space-alien:
Thanks Ben.

I've made a simplification to one of the unit tests. See what you think? This could be applied to the others.

Yes, we should add TapIf signatures.

PR #192, opened 2020-03-24
TapIf overloads for handling result-returning funcs

benprime:
TapIf overloads to handle result-returning funcs and associated tests.

@space-alien @vkhorikov  These should include all the related overloads for TapIf


vkhorikov:
Merged. Will publish a new version soon.

Issue #194, opened 2020-04-03
Better Debugging Experience


VladimirMakaev:
Hi I'm having great time using this library. One thing I've noticed as a potential enhancement is to either have ```ToString()``` method overridden or have a ```[DebuggerTypeProxy(...)]``` attribute applied to ```Result<...>``` types. That would make debugging much more pleasant.

Would you be interested in contribution in this area? Which approach would you like to follow?

vkhorikov:
I like the idea and I think we already have `ToString()` overridden in `Maybe`. How would you like to override `ToString()` in `Result` ?

VladimirMakaev:
Conceptually something like that
for Result<T,E>:
```c#
        public override string ToString()
        {
            return IsSuccess ? $"Success({Value})" : $"Failure({Error})";
        }
```
for Result<T>

```c#
        public override string ToString()
        {
            return IsSuccess ? $"Success({Value})" : $"Failure";
        }
```
I can performance optimize it to get rid of allocations on each call if it's OK conceptually.

vkhorikov:
Merged the PR. The optimization is optional. If it reduces readability, I would rather keep the code as it is.

PR #195, opened 2020-04-06
Implement ToString() for Result<T>, Result<T,E>, Result

VladimirMakaev:
#194 

Implement ToString() for Result<T>, Result<T,E>, Result to make the debugging experience more pleasant for the clients of this library


vkhorikov:
Looks great. Merged, thanks.

PR #212, opened 2020-07-09
Add Maybe<T>.ToList()

svroonland:
This is a useful operator, present in Scala as well for example.


hankovich:
It would be nice to have unit tests for this method.

svroonland:
@hankovich Done!

PR #214, opened 2020-07-14
Add async overloads for Maybe's Where, Map and Bind

hankovich:
It's not very comfortable to use maybe in async context since there're almost no extensions for `Task<Maybe<T>>`. This PR adds async overloads for Where, Map and Bind as well as tests for them.


vkhorikov:
Published in 2.11.0

PR #217, opened 2020-07-23
Fix Maybe.TryFirst with predicate for types where null is not the default

svroonland:
It would return a Maybe with Value for any type for which type default is not 0, eg primitives, tuples, even though there was no item in the enumerable matching the predicate.


vkhorikov:
Nice one. Merged, thanks

PR #219, opened 2020-07-31
Added Aggregate function for Result array

linkdotnet:
This will introduce a new extension method: `Aggregate`.

Sometimes you will run in the following situation:
```csharp
private static Result<MyObject> CreateMyObjects(IList<CreateParameter> createParameter)
{
    var results = createParameter.Select(CreateMyObject);
```

This will give you a list / enumeration of results. But for later uses you just wanna know if one of those have failed or not.

This where `Aggregate` comes in handy:
```csharp
private static Result<MyObject> CreateMyObjects(IList<CreateParameter> createParameter)
{
    var results = createParameter.Select(CreateMyObject).Aggregate();
    if (results.IsFailure)
```

### DIFF
--- a/CSharpFunctionalExtensions/Result/Extensions/Aggregate.cs
+++ b/CSharpFunctionalExtensions/Result/Extensions/Aggregate.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace CSharpFunctionalExtensions
+{
+    public static partial class ResultExtensions
+    {
+        /// <summary>
+        ///     Creates a new result from a enumeration of results. If one fails the new result will have the combined result
+        /// </summary>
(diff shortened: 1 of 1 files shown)


linkdotnet:
@vkhorikov Currently the signature is the following:
```csharp
public static Result<IEnumerable<TValue>> Aggregate<TValue>(this IEnumerable<Result<TValue>> results ...
```
I am still not 100% sure if we should return `Result<IEnumerable<TValue>>` or something like `Result<IReadOnlyCollection<TValue>>`.

Do you have a strong opinion on that?

hankovich:
Hi, @linkdotnet 
Could you please explain how `Aggregate` differs from the existing [Combine](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Result/Extensions/Combine.cs)?

linkdotnet:
Hi @hankovich 
that's a fair point. Didn't see that though. Of course with `Combine` this is obsolete. Thank you very much. I will close the PR.

PR #223, opened 2020-08-04
Remove unnecessary generic constraints and add tests

hankovich:
`ToResult` extensions for `Task<Maybe<T>>` have `T: class` generic constraint (I believe because of historical reasons). This PR removes them and add tests for these extensions (they were not covered previously).


vkhorikov:
Merged, thanks. The restriction was probably from the times when `Maybe` only worked with reference types.

PR #232, opened 2020-09-25
Fixes #183

### LINKED ISSUE #183: OnSuccessTry doesn't handle async lambda
I've got a pre-condition followed by a try that I am translating to chained results:

``` CSharp
return await Result
    .SuccessIf(condition, "Condition error message")
    .OnSuccessTry(async () => await ThingThatMightFail, "Fail message");
```

On closing the expression Intellisense lights up saying it can't await type of `Result<Task>`. Seems like `OnSuccessTry` is missing the magic that made this work with other success methods.

`OnSuccessTry` also doesn't fit well with the new method naming scheme. Maybe change it to `TryMap` or something?

In the mean time I had to factor out the chaining:

``` CSharp
var result = Result.SuccessIf(condition, "Condition error message");
if (result.IsSuccess)
    result = await Result.Try(async () => await ThingThatMightFail, "Fail message");
return result;
```

Note that `await Try(async () => foo,` works here.

kakone:
Adds extension methods with `Func<Task>` as parameter to fix #183.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/ExtensionTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/ExtensionTests.cs
@@ -129,7 +129,7 @@ namespace CSharpFunctionalExtensions.Tests.ResultTests
         {
             var originalResult = Result.Success();
 
-            var result = originalResult.OnSuccessTry(() => throw new Exception("execute action exception."));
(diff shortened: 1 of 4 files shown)


vkhorikov:
Merged, thank you. Will be available in v2.11.8.

PR #257, opened 2021-01-09
Changed errorMessage to errorPredicate for Ensure and added errorPred…

darickc:
…icate to async Ensure methods where object is available to generate error message.


vkhorikov:
Merged, thanks. Will be published soon.

PR #260, opened 2021-02-11
fix: Ensure with errorPredicate checks initial result state

draptik:
This patch to the Ensure methods adds a missing check to see if the input result is in failure state.
Also added some tests to confirm the behavior.


vkhorikov:
Merged, thanks. Will be published in 2.14.2.

Issue #264, opened 2021-02-17
Maybe<T>.None == null returns false


petrechitashvili:
Consider the following:

```csharp
Maybe<Class> maybeClassFromEmpty = null;
	
Console.WriteLine(maybeClassFromEmpty == null); // returns false
```
Considering that `Maybe<T>` implicitly converts `null` to `Maybe<T>.None` I'd expect comparing it with null to return `true`

pfitoussi:
Considering that Maybe<T> is a struct, it would never be null, therefore your statement is useless and you should instead write : 
`Console.WriteLine(maybeClassFromEmpty.HasNoValue)`

The `Maybe<T>` is a wrapper around a `T` (or `null`, `default`) and expose properties to handle both case: HasValue and HasNoValue.

I recommend you to read this article [Functional C#: Non-nullable reference types](https://enterprisecraftsmanship.com/posts/functional-c-non-nullable-reference-types/).

The roslynator extension warn us about the [RCS1166: Value type object is never equal to null](https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1166.md) when you write:
`maybeClassFromEmpty == null`.

petrechitashvili:
> therefore your statement is useless

My statement may be useless, I'll take that, but it is a legal statement and it leads to subtle issues.

We are modifying parts of our codebase with Maybe & Result types, not all at once and we don't (yet) have the luxury of checking everything with Roslynator. 

I'd expect that `Maybe<T>.None == null` returns true or throws an exception, but not false.

hankovich:
@pfitoussi 
>Maybe is a struct, it would never be null

That's not right in 100% cases. You're right that value type could never be null, however value type can provide it's own implemetation of `==` (which can return true).

Let me summarize:
```csharp
MyStruct value = ...
var result = value == null; // can return true, depends on == operator
var result = value is null; // will never return true, calls object.ReferenceEquals(object, object) under the hood, i.e. calls ReferenceEquals((object)value, null) and boxed struct will never be null
```

vkhorikov:
An additional logic in `==` wouldn't hurt, indeed.

petrechitashvili:
Thank you!

PR #267, opened 2021-02-18
Solution for issue #264

renato04:
Issue #264

Compare Maybe<T> with null value using `==` operator returns `true`

### DIFF
--- a/CSharpFunctionalExtensions/Maybe/Maybe.cs
+++ b/CSharpFunctionalExtensions/Maybe/Maybe.cs
@@ -48,7 +48,10 @@ namespace CSharpFunctionalExtensions
                 return maybe.Equals(value);
 
             if (maybe.HasNoValue)
-                return false;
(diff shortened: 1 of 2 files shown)


petrechitashvili (approved):
Looks good to me! 🙂

renato04:
Ops, sorry for the miss click

CSharpFunctionalExtensions/Maybe/Maybe.cs
```diff
@@ -48,7 +48,10 @@ public static Maybe<T> From(T obj)
                 return maybe.Equals(value);
 
             if (maybe.HasNoValue)
-                return false;
+                if (ReferenceEquals(value, null))
```
hankovich, line 51: can be just 
```csharp
if (maybe.HasNoValue)
    return value is null;
```

vkhorikov:
Merging this, will be published soon.

Issue #259, opened 2021-02-04
Proposal: additional MapError overloads


hankovich:
As I see there's no way to fluently convert `Result<T, E>` to `Result<T>`.
So I think the following overloads would be handly.

```csharp
public static Result<T> MapError<T, E>(this Result<T, E> result, Func<E, string> errorFactory);
public static Task<Result<T>> MapError<T, E>(this Task<Result<T, E>> resultTask, Func<E, string> errorFactory);
```

vkhorikov:
Sounds good to me!

PR #271, opened 2021-02-27
Additional MapError overloads

hankovich:
Implements #259


vkhorikov:
Merging this, will be published in v2.14.5

PR #277, opened 2021-03-15
add an overload to avoid nesting

michaeldileo:
Just ran into a situation where `.Map` returned a `Task<Result<T>>` and the return type was `Result<Result<T>>`. This will allow the return type to be `Result<T>`.

Example Code
``` csharp
Result<Y> result = await Maybe<X>.From(something)
    .ToResult("My Message")
    .Map(async x => 
    {
        Result<Y> y = await GetYAsync();
        y.OnFailure(log);
        // something else
        return y; // formerly this would result in Result<Result<Y>>
    });
```


vkhorikov:
Merging, thank you. Will publish with v2.15.1.

PR #282, opened 2021-04-22
Success and Failure async methods

fehimdervisbegovic:
Async Success and Failure method. We use these methods in our Orleans projects very intensively. So they would be very useful to us and probably for the other developers.


vkhorikov:
Thanks, appreciate the contribution. Could you describe the use case? These methods look like sheer async wrappers on top of the sync versions.

SuperJMN:
@fehimdervisbegovic Could you, please, provide some usage scenario to understand its usefulness? Thanks :)

PR #284, opened 2021-04-26
Proposal to use static helpers for shared logic in ResultCommonLogic.

stiankroknes:


vkhorikov:
Nice refactoring, thanks!

PR #285, opened 2021-05-06
Add overload to Result<T,E>.Ensure() for an error predicate

SamuelViesselman:
Example use case:

```
var result = GetCustomerWithId(id)
    .Ensure(c => c.isActive, c => new BadRequest($"Customer {c.name} is not active"));
```


vkhorikov:
Merged, thanks. Will be published soon with v2.15.2

Issue #294, opened 2021-06-14
Recommendation: Maybe<T>.IfNone()


seangwright:
`Maybe<T>.UnWrap()`and `Maybe<T>.Match` are great if you want to leave the monad and operate on values, but a lot of the benefits of the monad come with the fluent chaining of operations and transformations.

Often, as a `Maybe<T>` is working its way through layers of an application, there comes a point at which a default/fallback value is known and can be provided to the `Maybe<T>` but I don't what to unwrap it.

Example:

```csharp
Maybe<string> firstName = "sean";
Maybe<string> lastName = "wright";

// How do I supply a fallback value (in the case of `Maybe<string>.None` without unwrapping, like "unknown"?
Maybe<string> fullName = firstName.Bind(f => lastName.Map(l => $"{f} {l}"));

// later on ...

// if `fullName` was empty earlier I'd like it to have had the value "unknown" so that `uniqueFullName`
// wouldn't be just a Guid
Maybe<string> uniqueFullName = fullName.Map(n => $"{n}{Guid.NewGuid()}");
```

I propose something like `IfNone()`:

```csharp
public static Maybe<T> IfNone<T>(this Maybe<T> source, T fallback)
{
    if (source.HasNoValue)
    {
        return fallback;
    }

    return source;
}
```

Or maybe it would make sense to add an overload to `Maybe<T>.Map()` that took a 2nd parameter of a fallback value?

vkhorikov:
This is a good idea, but I'm not sure about the name of the method. I saw other libraries named it something like `OrElse`.

How about `Or`? Also, there could be 2 overloads, one for  a `T` argument and the other one for `Maybe<T>`.

As a side note, in your example, it looks like the full name will contain `None` if the first name is empty, but the last name is not. I would expect the full name to contain either both names, or one of the names if the other one is empty.

seangwright:
> 
> 
> This is a good idea, but I'm not sure about the name of the method. I saw other libraries named it something like `OrElse`.
> 
> How about `Or`? Also, there could be 2 overloads, one for a `T` argument and the other one for `Maybe<T>`.

Is this what you're thinking?

```csharp
public static Maybe<T> Or<T>(this Maybe<T> source, T value);
```

Would the `Or` that takes a `Maybe<T>` confuse with the `Bind()` semantics? Should it include `Bind*` in the name or is it best to keep this concepts different?

Naming is hard!

> 
[...]

PR #296, opened 2021-06-15
Add Maybe.Or sync/async extensions and tests

seangwright:
Implementation for https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/294

I added these:

```csharp
public static Maybe<T> Or<T>(this Maybe<T> maybe, T value)
public static async Task<Maybe<T>> Or<T>(this Task<Maybe<T>> maybeTask, T value)
```

Do we want any additional overloads here?

Ex (including async equivalents):

```csharp
public static Maybe<T> Or<T>(this Maybe<T> maybe, Maybe<T> fallback)
public static Maybe<T> Or<T>(this Maybe<T> maybe, Func<T> fallbackOperation)
public static Maybe<T> Or<T>(this Maybe<T> maybe, Func<Maybe<T>> fallbackOperation)
```

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
@@ -372,6 +372,32 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             maybe2.Value.Should().Be(result);
         }
 
+        [Fact]
+        public async Task Async_Or_returns_source_if_source_has_value()
+        {
+            var instance = new MyClass { Property = "Some value" };
+            Task<Maybe<MyClass>> maybeTask = GetMaybeTask(instance);
+            var result = new MyClass { Property = "Different value" };
+
+            Maybe<MyClass> maybe2 = await maybeTask.Or(result);
+
+            maybe2.HasValue.Should().BeTrue();
+            maybe2.Should().Be(instance);
+        }
+
+        [Fact]
+        public async Task Async_Or_returns_a_new_instance_with_value_when_source_is_empty()
+        {
+            var instance = Maybe<MyClass>.None;
+            Task<Maybe<MyClass>> maybeTask = GetMaybeTask(instance);
+            var result = new MyClass { Property = "Different value" };
+
+            Maybe<MyClass> maybe2 = await maybeTask.Or(result);
+
+            maybe2.HasValue.Should().BeTrue();
+            maybe2.Should().Be(result);
+        }
+
         private static Task<Maybe<MyClass>> GetMaybeTask(Maybe<MyClass> maybe) => maybe.AsTask();
 
         private static Func<MyClass, T> ExpectAndReturn<T>(MyClass expected, T result) => actual =>
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
@@ -373,7 +373,7 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
         }
 
         [Fact]
-        public async Task Async_Or_returns_source_if_source_has_value()
+        public async Task Async_Or_fallback_value_returns_source_if_source_has_value()
         {
             var instance = new MyClass { Property = "Some value" };
             Task<Maybe<MyClass>> maybeTask = GetMaybeTask(instance);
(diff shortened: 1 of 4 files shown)


vkhorikov:
> Ex (including async equivalents):

Yep, those overloads look useful too.

seangwright:
Updated with all the overloads!

vkhorikov:
Merged, thanks! Will publish a new version tomorrow.

hankovich:
Am I correct that `Maybe<T> Or<T>(this Maybe<T> maybe, T value)` will only return `None` when `value is null`?
If so, I'd suggest to change the return type to `T` (i.e. `T Or<T>(this Maybe<T> maybe, T value)`) and (optionally) introduce another overload:`Maybe<T> Or<T>(this Maybe<T> maybe, Maybe<T> value)`

@vkhorikov what's your opinion on this?

seangwright:
> Am I correct that Maybe<T> Or<T>(this Maybe<T> maybe, T value) will only return None when value is null?

It will return `Maybe<T>.From(T value)` when `maybe.HasNoValue` is true.

> If so, I'd suggest to change the return type to T

The point of this set of `.Or()` methods was to continue the `Maybe` but have a complimentary set of operations to `.Map()` that are called only when the calling `Maybe<T>` has no value, not necessarily to unwrap since we already have that with `.UnWrap()`.

> and (optionally) introduce another overload:Maybe<T> Or<T>(this Maybe<T> maybe, Maybe<T> value)

@hankovich do you mean [this overload](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs#L140)?

hankovich:
And `Maybe<T>.From(T value)` will only return `None` if `value` is null :)

So returning Maybe from `Maybe<T> Or<T>(this Maybe<T> maybe, T value)` only makes sence if `value` could be null. But value should never be null, because there's no point to use both Maybe and have nulls in your variables, right? That's why I think `Maybe<T> Or<T>(this Maybe<T> maybe, T value)` should return `T`, not `Maybe<T>`. And, as you rightly noted, this partucular overload of `Or` becomes extremely similar to `Unwrap`.

Yes, I meant exactly that overload.

seangwright:
That makes sense! 😀

I was thinking in terms of being able to continue to `.Bind()` ... if we leave the `Maybe<T>` we have to manually jump back in to be able to have expressions instead of statements.

```csharp
Maybe<string> maybe = //

maybe
    .Map(s => $"{s} test")
    .Or("test")
    .Bind(sTest => maybeReturningMethod().Map(sNew => $"{sNew} {sTest}");
```

With this `Or()` I don't have to go back to statements if I still have more transformations to complete.

I'm ok with removing it or adding a comment that its the same as `Maybe<T>.From(maybe.UnWrap(T fallback))`

hankovich:
You can remove `Maybe<T> Or<T>(this Maybe<T> maybe, T value)` (because it's very similar to Unwrap), but leave `Maybe<T> Or<T>(this Maybe<T> maybe, Maybe<T> value)`.

`Maybe<T>` has an implicit conversion from `T` to `Maybe<T>`, so your chain

```csharp
maybe
    .Map(s => $"{s} test") // returns Maybe<string>
    .Or("test") // converts string to Maybe<string> and calls Maybe<string> Or(Maybe<string>, Maybe<string>)
```
will work exaclty as you expected!

hankovich:
btw do you think the name `Or` is clear enough? It took me a while to figure out what method is doing. At first I decided it's something about boolean logic :)
What do you think about plain old `Coalesce`?

vkhorikov:
Great point regarding the `Maybe<T> Or<T>(this Maybe<T> maybe, T value)` overload. Looks like we don't need it indeed.

Regarding the name of the `Or` method. I don't quite like `Coalesce`. Its behavior is more akin to `Unwrap` where the method changes the type from `Maybe<T>` to `T`. Example:

```csharp
int? variable1 = null;
int variable2  = variable1 ?? 4; // The coalesce operator changes Nullable<int> to int
```

There probably is an FP (monadic) term for this function, but I'm familiar with it.

Thoughts?

hankovich:
It depends. The null-coalescing operator can also be used between nullable types:

```csharp
int? variable1 = null;
int? variable2 = null;
int? variable3 = variable1 ?? variable2;
```

I checked `option`/`maybe` types in Haskell, Scala, OCaml, List, F# and even Java. Some of them do not have the corresponding functionality, but those packages that do always use the name `orElse` (Scala, F#, Java).

seangwright:
I like going with any established patterns - `OrElse<T>(...)` is definitely readable.

vkhorikov:
Yes, `OrElse` is what I saw in other libraries too. Let's do a small voting here. Do you guys both prefer `OrElse` over `Or`?

I like `Or` a little bit better, but don't have a strong opinion. If you both prefer `OrElse`, I'm fine with that.

seangwright:
I have no preference either!

What's the general direction of the library? Is it a way to introduce C# devs to functional ideas, or match functional terminology, or be somewhere in the middle?

The goal of LanguageExt is pretty obvious :) but I use this project because it lets me mix functional and procedural logic without too much pain. It lets me introduce devs to functional programming gradually without being overwhelming. I've found this second point to be important.

vkhorikov:
The goal is to provide functional programming features without much diversion from the C# canon. Hence all the little tricks e.g. implicit conversion from `null` to `Maybe`, so that the resulting code doesn't look much different from the usual way of doing this in C#.

@hankovich what's your preference of `Or` vs `OrElse`?

hankovich:
I think I like `Or` better. In my opinion, the `Else` suffix doesn't really matter here.
This is very Java style to add additional "explanatory" suffixes (Java's `optional` also has such methods as `ifPresentOrElse` and `orElseGet`).

seangwright:
I'll make a PR to remove that extra `Maybe<T>.Or()` and we'll be good to go!

vkhorikov:
> I'll make a PR to remove that extra Maybe<T>.Or() and we'll be good to go!

Sounds good! Let's keep `Or`.

PR #297, opened 2021-06-19
Remove redundant Maybe.Or extension, add Maybe.Or null test

seangwright:
Follow up for #296

Added a `null` test for `Or()`.


vkhorikov:
Publishing new v2.16.0

Issue #307, opened 2021-07-13
Possilbe issues when using CombineAsyncLeft if the IEnumerable<Task<Result> must be iterated in order 


YudApps:
[Entity Framework Core does not support multiple parallel operations being run on the same DbContext instance. ](https://docs.microsoft.com/en-us/ef/core/dbcontext-configuration/#avoiding-dbcontext-threading-issues).

This mean that a code like:
``` CSharp
ids.Select(async () => Result.Try(await dbContext.DataSet.FirstOrDefault(e => e.Id = id))
     .Combine()
```
May result with an error:

>  second operation was started on this context before a previous operation completed. This is usually caused by different threads concurrently using the same instance of DbContext. For more information on how to avoid threading issues with DbContext, see https://go.microsoft.com/fwlink/?linkid=2097913.

I think we should have an option:
``` CSharp

public enum AsyncExecutionOrder { Unordered = 0 , InOrder }

public static async Task<Result<IEnumerable<T>>> Combine<T>(this IEnumerable<Task<Result<T>>> tasks, string errorMessageSeparator = null, AsyncExecutionOrder  order= TaskCombineOrder.Unordered)
{
    Result<T>[] results = (order == AsyncExecutionOrder .Unordered)
                                      ? await Task.WhenAll(tasks).DefaultAwait();
                                      : await tasks.ExecuteInOrder()
    return results.Combine(errorMessageSeparator);
}

public static async Task<Result<IEnumerable<T>>> ExecuteInOrder<T>(this IEnumerable<Task<Result<T>>> tasks)
{
    List<Result<T>> results = new List<Result<T>>();
    foreach (var task in tasks)
    {
        results.Add(await task);
    }
    return results;
}
```

I tested it on my case with:
``` CSharp
        public static async Task<Result<IEnumerable<T>>> CombineInOrder<T>(this IEnumerable<Task<Result<T>>> tasks, string? errorMessageSeparator = null)
        {
            List<Result<T>> results = new List<Result<T>>();
            foreach (var task in tasks)
            {
                results.Add(await task);
            }
            return results.Combine(errorMessageSeparator);
        }
```
and it fix the async issue of the DbContext.

I can implement it if this suggestion will be approved.

vkhorikov:
We can add the second method (`CombineInOrder`), but I don't think the `AsyncExecutionOrder` enum and the additional argument in `Combine` are needed.

YudApps:
[...]

PR #309, opened 2021-07-18
CombineInOrder IEnumerable<Task<Result<T>>> extension

YudApps:
#307 Create and test CombineInOrder extension method that keep tasks execution order and avoid EF.Core aysnc issues: https://go.microsoft.com/fwlink/?linkid=2097913.


vkhorikov:
Great! Will be published soon with v2.18.0

PR #315, opened 2021-08-12
Fix Result.Ensure

seangwright:
There are 2 overloads of `Result.Ensure` that execute the error predicate even when the 'ensure' predicate passes.

This can lead to unexpected behavior and exceptions because the error predicate might access the `T` value in a way that the failing predicate would imply is valid.

For example, if `Result<User>.Ensure(user => user is null, user => $"User {user.Email} is not null");` passes and the error predicate is still evaluated, it will lead to a `NullReferenceException`.

The error predicate should only be evaluated when the predicate fails.

I added 2 tests to verify the changes.


vkhorikov:
Merging, will publish soon.

PR #318, opened 2021-08-17
Add non-generic Maybe.None

### LINKED ISSUE #317: Non-generic Maybe.None
Since there's now a non-generic `Maybe` class, it would be nice to have a `Maybe.None` member which could implicitly cast to a `Maybe<T>`

```csharp
public Maybe<int> GetInvalidValue()
{
    return Maybe.None;
}
```

For this change `Maybe` class will need to be changed to be non-static. Maybe change it to `struct` also, to make it consistent with `Maybe<T>`.

I can send a PR if you think this might be a good idea.

Razenpok:
Closes #317


vkhorikov:
Looks great, merging this. Will be published with v2.20.0 shortly.

PR #320, opened 2021-08-24
Added test for comparing two value objects with different type

AliKhalili:
Hi
Thank you for your great repository, your contribution and valuable book in Unit Testing
There is not any unit test when comparing two different value objects which have a different type. In this situation, `CompareTo`  try to comparing string representation of value object type together.
It seems this new test is a trivial test that I added to the Test Project but it might be helpful for those that want to change this comparison implementation(using value comparison instead of string representation of value object type).


vkhorikov:
Yep, good test, merging it, thank you.

Issue #293, opened 2021-06-10
Missing Maybe.Execute(Func<T, Task> action)


SuperJMN:
Hi! Congrats on the library!

I'm missing a method like the one on the subject line to be able to await a `Task`.

See the example:

```
                await toDelete.Execute(async comment => 
                {
                    context.Comments.Remove(comment);
                    await context.SaveChangesAsync();
                });
```

vkhorikov:
Yep, looks like an overload is missing in `AsyncMaybeExtensions.cs`

PR #326, opened 2021-09-16
Feature: Maybe async Execute

seangwright:
Handles #293.


vkhorikov:
Merged, will be published soon.

Issue #324, opened 2021-09-10
Feature: Extensions for IEnumerable<Maybe<T>>


seangwright:
I was wondering if there was any value in extensions like these:

```csharp
public static IEnumerable<T> MapValues<T, K>(this IEnumerable<Maybe<T>> maybes) =>
    maybes.Where(m => m.HasValue).Select(m => m.Value);

public static IEnumerable<K> MapValues<T, K>(this IEnumerable<Maybe<T>> maybes, Func<T, K> selector) =>
    maybes.Where(m => m.HasValue).Select(m => selector(m.Value));
```

I do run into situations where I have a collection of `Maybe<T>` and I want to filter to only the items in the collection that have a value and then map that value.

Currently, I use the implementation `maybes.Where(m => m.HasValue).Select(m => m.Value)`, and even though there's no other way to do it, I'd prefer not to see it in my code :)

I'm not sold on the name `.MapValues()`, but I'm not sure what would be appropriate.

vkhorikov:
Yeah, I use `maybes.Where().Select()` a lot too. A new extension method wouldn't hurt. Regarding naming, `FilterValues` would probably be better. The method contains both filtration (`Where`) and mapping (`Select`), but I think the emphasis here should be on filtration, the mapping part is going to be obvious from the signature. I'm open for other suggestions too, though.

Razenpok:
A new extension method would be great indeed, through you can currently use `.Choose(x => x)` to compress `.Where(x => x.HasValue).Select(x => x.Value)`

vkhorikov:
Wait, so `Choose` is exactly this method then. We can add an overload without the `selector` argument and probably probably refactor it, to replace `yeild` with the regular LINQ.

We can rename it too (and keep the old method with an `[Obsolete]` mark). What's your opinion on the method name @Razenpok @seangwright ?

* `Choose`,
* `FilterValues`, or
* `MapValues` ?

Razenpok:
At first, I was like

> I think `FilterValues()` and `MapValues(selector)` will make the most sense. Not even sure why `Choose` was picked as a name for such an operation.

But then

> Oh, so there's a `choose` list operator in F# and that's why we have `Choose` here.

This opens up a discussion of the target audience of the library and which side does this library incline to:

- On the one side is F# mimicking: going with `ChooseValues()` and `Choose(selector)`
- On the other side is C# mimicking: going with `WhereHasValues()` and `SelectValues(selector)`
[...]

PR #327, opened 2021-09-18
Documentation Updates / Maybe.Choose overload - WIP

seangwright:
This is in response to #324 and includes the overload to `Maybe<T>.Choose()` and Work-in-progress updates to the `README.md`.

The general idea is to add API examples of the library with brief explanations of when and how to use each of the features. This would be something useful to developers that are new to and experienced with the library.

I've added a new section to the README and begun the `Maybe` examples.

I'm now realizing just how many examples there will be with all the methods and overloads. Double that with the `Task` ones!

Maybe this should be in an `EXAMPLES.md` that we link to from the `README.md`?

I know there's also the `CSharpFunctionalExtensions.Examples` project that has a couple `Result` examples and expanding that is also an option. I find the Markdown to be more immediately useful and since the library rarely has breaking changes, being able to have refactorings with the examples isn't a huge benefit.

Thoughts?

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,10 @@
+{
+    "line-length": {
+        "line_length": 120
+    },
+    "code-block-style": {
+        "style": "fenced"
+    },
+    "code": true,
+    "code-fence-style": true
+}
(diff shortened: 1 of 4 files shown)


vkhorikov:
Looks very good. I prefer an `md` file as well, `CSharpFunctionalExtensions.Examples` are there to ensure backward compatibility of new changes with the older usage patterns, they weren't meant (at least initially) as a replacement for the documentation.

Note that implicit conversions for `Maybe` can also do this:

```cs
Maybe<string> myString = null; // works when returning a value from a method as well
```

I personally prefer this syntax to `Maybe.None<string>` because of brevity.

Razenpok:
@vkhorikov FYI you can use `default` instead of `null` which also works for `Maybe<T>` where `T` is a value type

vkhorikov:
@Razenpok Indeed!

Issue #328, opened 2021-09-20
Feature: Maybe.Execute() when there is no value


seangwright:
`Maybe.Execute()` is effectively shorthand for:

```csharp
Maybe<int> maybe = 10;

maybe.Match(
    number => Console.WriteLine(number),
    () => {});
```

I occasionally come across situations where I want the reverse:

```csharp
Maybe<int> maybe = Maybe<int>.None;

maybe.Match(
    _ => {},
    () => Console.WriteLine("There's nothing here"));
```

Can we add an overload to `Maybe.Execute` that takes an `Action` instead of an `Action<T>` and would be called when there is no value?

```csharp
Maybe<int> maybe = Maybe<int>.None;

maybe.Execute(() => Console.WriteLine("There's nothing here"));
```

Razenpok:
I'm afraid such operator would cause confusion. Maybe `Compensate` would be a better one?

vkhorikov:
I agree regarding confusion, but not sure about `Compensate`. What about `ExecuteNoValue` ?

PR #329, opened 2021-09-21
Feature: Maybe.ExecuteNoValue, Maybe.Execute async action, update README examples

seangwright:
This PR:
- Handles #328 for `Maybe.ExecuteNoValue`.
- Adds the ability to pass `Task` returning 'actions' to `Execute` (and `ExecuteNoValue`)
- Updates the README where I noticed some of the examples weren't correct
- Adds the beginning of the `Result` examples

### DIFF
--- a/CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs
+++ b/CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace CSharpFunctionalExtensions
 {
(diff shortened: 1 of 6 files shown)


CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs
```diff
@@ -100,6 +107,48 @@ public static void Execute<T>(this Maybe<T> maybe, Action<T> action)
...
+            action();
+        }
+        
+        /// <summary>
+        /// Executes the given async <paramref name="action" /> if the <paramref name="maybe" /> has a value
+        /// </summary>
+        /// <param name="maybe"></param>
+        /// <param name="action"></param>
+        /// <typeparam name="T"></typeparam>
+        public static async Task Execute<T>(this Maybe<T> maybe, Func<T, Task> action)
```
seangwright, line 130: While these are async methods they work with 'sync' `Maybe`s, so I put them in this file instead of `AsyncMaybeExtensions`. `.Match` is in here even though it could return a `Task`.

But, I can move them if you think that makes sense.

README.md
```diff
@@ -99,6 +99,14 @@ int storeInventory = ...
 Maybe<string> fruit = storeInventory > 0
     ? Maybe<string>.From("apple")
     : Maybe<string>.None;
+
+// or where the generic type is a reference type
+
+Maybe<string> fruit = null;
```
seangwright, line 105: I added your 'none' variation examples

README.md
```diff
@@ -224,7 +232,7 @@ Maybe<string> apple = "apple";
 
 Maybe<string> favoriteFruit = apple.Where(IsMyFavorite);
 
-Console.WriteLine(favoriteFruit); // "No value"
+Console.WriteLine(favoriteFruit.ToString()); // "No value"
```
seangwright, line 235: I realized these would all throw if the `Maybe` is 'none' and we don't explicitly call `.ToString()`.

I'm used to always embedding my `Maybe` values in interpolated strings or using `Execute` to get the value vs printing the `Maybe` itself.

CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs
vkhorikov: No need, looks good as-is

vkhorikov:
Everything looks very good. There seem to be some conflicts with the master branch. Could you resolve them?

seangwright:
All resolved!

vkhorikov:
Thanks much! Merging, will be published as v2.20.4 shortly

PR #331, opened 2021-09-27
Add Maybe.GetValueOrThrow() and Maybe.GetValueOrDefault()

### LINKED ISSUE #312: GetValueOrDefault?
I'm searching the Result type for something like below but I cant find this functionality. Is there any way to invoke Result<T> type with something like GetValueOrDefault(T default) so when it is success it return it internal value but on error it return given default?

### LINKED ISSUE #330: Overload for Option.Unwrap that takes a function
Let's say we have the following situation:
```
 // Get parent's favorite child, returning None if they lied about being neutral :)
Maybe<Child> GetFavoriteChild(Parent parent) => ...

// Expensive computation to predict a likely favorite
Child PredictFavoriteChild() => ... 

// Option 1
Child WhoGetsTheLastCookie(Parent parent)
{
  var defaultChild = PredictFavoriteChild(); // Expensive computation is always performed
  return GetFavoriteChild(parent).Unwrap(defaultChild );
}

// Option 2 - currently works just fine, just not as clean as Option 3 (IMO)
Child WhoGetsTheLastCookie(Parent parent)
{
 var childOption = GetFavoriteChild(customer);
 return childOption .HasValue ? childOption .Value : PredictFavoriteChild();
}

// Option 3 - proposed Unwrap overload
Child WhoGetsTheLastCookie(Parent parent) =>
  GetFavoriteChild(customer)
    .Unwrap(PredictFavoriteChild);
```

In F#, there is a **defaultWith** function that looks like this:
```
Option.defaultWith : (unit -> 'T) -> ('T option):  'T
```

I believe this could fit well as an Unwrap overload:
```
public static T Unwrap<T>(this Maybe<T> maybe, Func<T> compensator)
```

I would find value in adding this overload, but was wondering whether it's generally useful enough to add to the library. Happy to create a PR if necessary.

Thanks!

Razenpok:
This is a draft of `Maybe.GetValueOrThrow()` and `Maybe.GetValueOrDefault()` replacement for `.Value` and `.Unwrap()`

If it looks good, I will also add `Result.GetValueOrThrow()`, `Result.GetValueOrDefault()`, `Result.GetErrorOrThrow()` and `Result.GetErrorOrDefault()`

Closes #312
Closes #330

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/ExtensionsTests.cs
@@ -10,42 +10,42 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
     public class ExtensionsTests
     {
         [Fact]
-        public void Unwrap_extracts_value_if_not_null()
+        public void GetValueOrDefault_extracts_value_if_not_null()
         {
             var instance = new MyClass();
             Maybe<MyClass> maybe = instance;
 
-            MyClass myClass = maybe.Unwrap();
+            MyClass myClass = maybe.GetValueOrDefault();
 
             myClass.Should().Be(instance);
         }
 
         [Fact]
-        public void Unwrap_extracts_null_if_no_value()
+        public void GetValueOrDefault_extracts_null_if_no_value()
         {
             Maybe<MyClass> maybe = null;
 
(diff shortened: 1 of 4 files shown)


vkhorikov:
Looks good! Merging this, will be published with v2.21.0

PR #332, opened 2021-09-29
Bind between Result<T, E> and UnitResult<E>

lqkahler:


vkhorikov:
Looks great, merging this, thanks you. Will be published shortly.

PR #334, opened 2021-10-04
Align EnumValueObject implementations

### LINKED ISSUE #286: Align EnumValueObject<TEnumeration, TId> and EnumValueObject<TEnumeration>
Introduction of `EnumValueObject<TEnumeration, TId>` is a great addition to the set of available classes in the library.
However, I see that implementations merged in #275 are not aligned.

I would like to have similar set of properties and methods available in `EnumValueObject<TEnumeration, TId>` :
1. `public static IEnumerable<TEnumeration> All = EnumerationsById.Values;`
2. `public static bool Is(string possibleName) => EnumerationsByName.ContainsKey(possibleName);`
3. `public static bool Is(int possibleId) => EnumerationsById.ContainsKey(possibleId);`

which are present in its counterpart `EnumValueObject<TEnumeration>`.

In addition to this alignment I would like to discuss two entries that could possibly benefit from refactoring:
1. Change `possibleKey` to `possibleId` in `public static bool Is(string possibleKey) => ...` in  `EnumValueObject<TEnumeration>` since `id` parameter is present in other method's signatures. This would make overall API consistent.
2. Change return type in `public static IEnumerable<TEnumeration> All = Enumerations.Values;` to `ICollection<TEnumeration>` to better express what's really returned (one can assume that IEnumerable content here is not materialized yet imho; refering [this pos](https://enterprisecraftsmanship.com/posts/return-the-most-specific-type/)t)
3. `ToString()` called on `EnumValueObject<TEnumeration, TId>` should return rather a `Name` than an `Id` (this is questionable since this behaviour will be somewhat different from what one can expect from `EnumValueObject<TEnumeration>` where `Id` is a `string`).

I can prepare a PR if you consider these changes valid.

MikelThief:
This PR closes #286.

@vkhorikov do you like they way the problem with missing `IReadOnlyCollection<>` is solved for .NET4.0 or do you prefer a different solution? This follows your suggestion from the issue (though I thought collection might be more appropriate than list).

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ValueObjectTests/EnumValueObjectTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ValueObjectTests/EnumValueObjectTests.cs
@@ -82,10 +82,10 @@ namespace CSharpFunctionalExtensions.Tests.ValueObjectTests
         [InlineData("Four", false)]
         [InlineData(nameof(TestEnumValueObject.Two), true)]
         [InlineData(nameof(TestEnumValueObject.One), true)]
-        public void GivenPossibleKey_WhenCheckingIfKeyIsEnumValueObject_ThenShouldReturnTrueIfKeyRecognized(string possibleKey, bool isIn)
+        public void GivenPossibleKey_WhenCheckingIfKeyIsEnumValueObject_ThenShouldReturnTrueIfKeyRecognized(string possibleId, bool isIn)
         {
             // Act
-            var isEnumValueObject = TestEnumValueObject.Is(possibleKey);
+            var isEnumValueObject = TestEnumValueObject.Is(possibleId);
 
             // Assert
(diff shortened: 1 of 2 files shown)


vkhorikov:
`IReadOnlyCollection` is fine too, but it's better to use `IEnumerable` instead of `ICollection` because `ICollection` exposes API that allows for collection mutation. We can leave it as is for now, though.

hankovich:
Nit: I would say it's better (at least for .NET 4.0) to call `ToArray` instead of `ToList` since it arrays are read-only collections.

```
ICollection<int> list = new List<int>(42);
ICollection<int> array = new int[42];

var listIsReadOnly = list.IsReadOnly; // false
var arrayIsReadOnly = array.IsReadOnly; // true
```

PR #336, opened 2021-10-05
Accept custom exceptions for GetValueOrThrow on Maybes

guythetechie:
Currently, ``GetValueOrThrow`` throws a generic ``InvalidOperationException`` when the maybe is none. These new extensions allow users to specify a custom exception that is more relevant to their application.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/AsyncExtensionsTests.cs
@@ -645,6 +645,23 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             instance.Value.Property.Should().Be("Some Value");
         }
 
+        [Fact]
+        public async Task Async_GetValueOrThrow_throws_requested_exception_if_none()
+        {
+            const string exceptionMessage = "Maybe is none";
+            var exception = new MyException(exceptionMessage);
+
+            Func<Task<int>> getMaybeValue = () =>
+            {
+                var maybe = Maybe<int>.None;
+                var maybeTask = Task.FromResult(maybe);
+                
+                return maybeTask.GetValueOrThrow(exception);
+            };
+
+            await getMaybeValue.Should().ThrowExactlyAsync<MyException>().WithMessage(exceptionMessage);
+        }
+
         private static Task<Maybe<MyClass>> GetMaybeTask(Maybe<MyClass> maybe) => maybe.AsTask();
 
         private static Func<MyClass, T> ExpectAndReturn<T>(MyClass expected, T result) => actual =>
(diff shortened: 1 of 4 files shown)


hankovich:
Thanks you for your contribution!
Could you please add more context? Which exception types are more appropriate to your projects?

As I know, `GetValueOrThrow` was added as a foolproof mechanism to prevent accessing `Value` without checking that maybe has value. And it uses `InvalidOperationException` to act like `Nullable<T>`.

```csharp
var maybe = Maybe<int>.None;

// 1
return maybe.GetValueOrThrow(); // looks fine for me

// 2
return maybe.GetValueOrThrow(new AbandonedMutexException()); // raises questions

// As for me, I would rewrite example #2 like this

// 3
int ThrowException()
{
    throw new AbandonedMutexException();
}

return maybe.GetValueOrDefault(ThrowException); // causes questions
```

vkhorikov:
Having a custom exception message would be beneficial, but not a custom exception type, that should remain the same. If you could modify the PR to accept a message instead of a whole exception, that would be good.

guythetechie:
Happy to modify the PR as requested.

If I understand correctly, you'd prefer:
```
// Gets inner value or throws new InvalidOperationException(errorMessage)
T GetValueOrThrow<T>(this Maybe<T> maybe, string errorMessage) => ...

// Example
var customer = TryGetCustomer(customerId, etag)
               .GetValueOrThrow("Could not find customer. Optimistic concurrency error.")
```

If I may ask though, why is that better than:
```
// Gets inner value or throws myException
T GetValueOrThrow<T>(this Maybe<T> maybe, Exception myException) => ...

// Example
var customer = TryGetCustomer(customerId, etag)
               .GetValueOrThrow(new OptimisticConcurrencyException())
```

To me, the latter is clearer and matches the function name well: we get the value or throw the specified exception.

Please note that I'm implementing this as an extension rather than an additional method on the ``Maybe`` struct. It's trivial to write though, and I certainly understand wanting to avoid polluting the library with extensions that don't offer widespread value.

vkhorikov:
> If I understand correctly, you'd prefer

That's correct.

> If I may ask though, why is that better than

This goes back to the philosophy of functional programming where you shouldn't use exceptions to control the program flow and should only use them for failing fast. The only reason to create custom exception types is to catch those exceptions, which is an anti-pattern in functional programming. More on this here: https://enterprisecraftsmanship.com/posts/custom-exception-types/

guythetechie:
> > If I understand correctly, you'd prefer
> 
> That's correct.
> 
> > If I may ask though, why is that better than
> 
> This goes back to the philosophy of functional programming where you shouldn't use exceptions to control the program flow and should only use them for failing fast. The only reason to create custom exception types is to catch those exceptions, which is an anti-pattern in functional programming. More on this here: https://enterprisecraftsmanship.com/posts/custom-exception-types/

Sounds good. I've updated the PR title and code accordingly.

vkhorikov:
Yeah, I agree. An instance method is more appropriate here. This should be an additional parameter in `GetValueOrThrow` with a default `null` value (see point 3 below).

@guythetechie could you please make a couple more changes?

1) Make the extension method an instance method instead
2) Replace the ternary operator with early exit
3) Introduce default exception message `Result.DefaultNoValueExceptionMessage` for situations when the `errorMessage` argument is `null`

PR #339, opened 2021-10-09
Add Finally Extension methods for UnitResult

SamuelViesselman:
Add tests for Finally Extension methods


hankovich:
Could you add more tests please?
It looks like all positive (`result.IsSuccess == true`) cases are fully covered, while there are no tests for failed results.

PR #340, opened 2021-10-10
Add Maybe.Deconstruct extension

Razenpok:
Probably supersedes #169


vkhorikov:
Well, I think a test for this particular feature would be nice, just to make sure that `Deconstruct` doesn't throw if the `Maybe` doesn't have a value.

PR #342, opened 2021-10-11
With Extensions (Combine)

SuperJMN:
These extensions will allow you combine Results.
This is done by combining successful values using a factory + merging errors using a factory

### Example

```
var r1 = Result.Success(1);
var r2 = Result.Success(3);
var combined = r1.With(r2, (x, y) => x + y);
```

combined will contain a success with a value of **4**

### DIFF
--- a/CSharpFunctionalExtensions/CSharpFunctionalExtensions.csproj
+++ b/CSharpFunctionalExtensions/CSharpFunctionalExtensions.csproj
@@ -36,6 +36,10 @@
   <ItemGroup Condition="'$(TargetFramework)'=='net40'">
     <PackageReference Include="Microsoft.Bcl.Async" Version="1.0.168" />
   </ItemGroup>
+	
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4')) OR $(TargetFramework)=='netstandard2.0'">
+    <PackageReference Include="System.ValueTuple" Version="4.*" />
+  </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0">
--- a/CSharpFunctionalExtensions/Result/Extensions/BindError.cs
+++ b/CSharpFunctionalExtensions/Result/Extensions/BindError.cs
@@ -1,0 +1,29 @@
+﻿using System;
+
+namespace CSharpFunctionalExtensions
+{
+    public partial class ResultExtensions
+    {
+        public static Result<T, E2> BindError<T, E, E2>(this Result<T, E> self,
+            Func<E, Result<T, E2>> func)
+        {
+            if (self.IsSuccess)
+            {
+                return Result.Success<T, E2>(self.Value);
+            }
+
+            return func(self.Error);
+        }
+
+        public static Result<T> BindError<T>(this Result<T> self,
+            Func<string, Result<T>> func)
+        {
+            if (self.IsSuccess)
+            {
+                return Result.Success(self.Value);
+            }
+
+            return func(self.Error);
+        }
+    }
+}
--- a/CSharpFunctionalExtensions/Result/Extensions/With/AsyncResultOfT.cs
+++ b/CSharpFunctionalExtensions/Result/Extensions/With/AsyncResultOfT.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace CSharpFunctionalExtensions
+{
+    public static partial class ResultExtensions
+    {
+        public static Task<Result<TResult>> WithMap<T1, T2, TResult>(this Result<T1> a,
+            Result<T2> b,
+            Func<T1, T2, Task<TResult>> func)
+        {
+            var mapSuccess =
+                a.BindError(e1 => b
+                        .MapError(e2 => Errors.Join(e1, e2))
+                        .Bind(_ => Result.Failure<T1>(e1)))
+                    .Bind(x => b
+                        .Map(y => func(x, y))
+                        .MapError(el => el));
+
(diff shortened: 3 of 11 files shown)


CSharpFunctionalExtensions/Result/Extensions/With/AsyncResultOfT.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace CSharpFunctionalExtensions
+{
+    public static partial class ResultExtensions
+    {
+        public static Task<Result<TResult>> WithMap<T1, T2, TResult>(this Result<T1> a,
```
hankovich, line 8: I feel it can be expressed via `SelectMany`

hankovich:
Personally I feel it's better to use `SelectMany` via the fluent syntax (which I hate and it's the single case where I'm ok to use it) for such complex cases.

And, speaking honestly, I think those extensions do too much. It's hard to understand all those positional and generic parameters.

Example from tests:
```csharp
await r1.WithMap(r2, (a, b) => Task.FromResult(a + b), (e1, e2) => "failure"); // I don't feel it's simple and easy to read/understand
```

CSharpFunctionalExtensions/Result/Extensions/BindError.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace CSharpFunctionalExtensions
+{
+    public partial class ResultExtensions
+    {
+        public static Result<T, E2> BindError<T, E, E2>(this Result<T, E> self,
```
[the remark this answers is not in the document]
SuperJMN: I don't think so. OnFailureCompensate expects the same error type :) This maps from `E`, to `E2`

CSharpFunctionalExtensions/Result/Extensions/With/AsyncResultOfT.cs
SuperJMN: If you're able to give an example, I'll be good with that :D

CSharpFunctionalExtensions/CSharpFunctionalExtensions.csproj
```diff
@@ -36,6 +36,10 @@
   <ItemGroup Condition="'$(TargetFramework)'=='net40'">
     <PackageReference Include="Microsoft.Bcl.Async" Version="1.0.168" />
   </ItemGroup>
+	
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4')) OR $(TargetFramework)=='netstandard2.0'">
+    <PackageReference Include="System.ValueTuple" Version="4.*" />
```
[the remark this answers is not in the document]
ProphetLamb: @SuperJMN Comparing `TargetFramework` is considered leaky, for various good reasons, such as Mono RT, and "oh I missed a version", additionally we are not interested in the `TargetFramework` at all, but in its features.
Features are expressed using preprocessor directives, so instead, compare using `DefineConstants.Contains('NET5_0_OR_GREATER')` which are supported by 3rd party runtimes and also documented in the dotnet documentation.

CSharpFunctionalExtensions/Result/Extensions/With/AsyncResultOfT.cs
```diff
@@ -0,0 +1,53 @@
...
+            return mapSuccess;
+        }
+
+        public static Task<Result> WithBind<T1, T2>(this Result<T1> a,
+            Result<T2> b,
+            Func<T1, T2, Task<Result>> map)
+        {
+            var mapSuccess =
+                a.BindError(el1 => b
+                        .MapError(el2 => string.Join(Result.ErrorMessagesSeparator, el1, el2))
```
ProphetLamb, line 44: Use `Errors.Join`

CSharpFunctionalExtensions/Result/Extensions/With/Errors.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CSharpFunctionalExtensions
+{
+    public static class Errors
+    {
+        public static string Join(string e1, string e2)
+        {
+            return string.Join(Result.ErrorMessagesSeparator, e1, e2);
```
ProphetLamb, line 7: Use string interpolation `$"{el1}{Result.ErrorMessagesSeparator}{el2}`

CSharpFunctionalExtensions/Result/Extensions/With/ResultOfT.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
+namespace CSharpFunctionalExtensions
+{
+    public static partial class ResultExtensions
+    {
+        public static Result<(T1, T2)> With<T1, T2>(this Result<T1> a, Result<T2> b)
```
ProphetLamb, line 7: Declare argument covariance `in this` & `in` for every non TAP, non closure usages accepting `struct`s.

CSharpFunctionalExtensions/Result/Extensions/With/ResultOfT.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            return a.WithMap(b, (x, y) => (x, y));
```
ProphetLamb, line 9: Use `static () {}` lambda in all instances where no closure allocation is required.

CSharpFunctionalExtensions/Result/Extensions/With/Errors.cs
ProphetLamb: Would be sensible to wait for #378 with the error combining

PR #349, opened 2021-10-14
Feature: add async fallback for synchronous Maybe.Or()

seangwright:
- New `Maybe.Or` overloads:

```csharp
Maybe<string> maybe = "test";

maybe.Or(() => Task.FromResult(Maybe.From("other"));
// and
maybe.Or(Task.FromResult("other"));
// and
maybe.Or(() => Task.FromResult("other"));
```

- Updated `README.md` to replace deprecated `Maybe` features `.Unwrap()` and `.Value`

Edit ---

Looks like VS Code did some auto formatting, removing whitespace, when I saved the file. I can undo that if needed.


seangwright:
As a side note, these `Maybe` test files are growing to be quite large.

Is it time to have a test file per-extension? Or event a file per maybe-extension method?

PR #355, opened 2021-11-01
Unit result convert failure

SamuelViesselman:
Added ConvertFailure methods for UnitResult

Also moved TestBase.cs up out of the Extensions folder so it could be used with more than just extension methods


vkhorikov:
Looks good, will be published shortly.

PR #363, opened 2021-11-16
And Compensate extensions and tests for them

hankovich:
Compensate (naming as per https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/312#issuecomment-939088642) is an improved version of legacy `OnFailureCompensate`.
The biggest issue with `OnFailureCompensate` (other than the long name) is the inability to change the type of  `TError`, and `Compensate` provides a rich set of overloads to easily both handle compensation and convert between different error types.


vkhorikov:
That's a huge chunk of work! Publishing all 4 PRs now with a new version.

PR #365, opened 2021-11-21
Target net6.0 and remove trimming warning

### LINKED ISSUE #364: IL2104 warning when publishing with `/p:PublishTrimmed=true` on .NET 6.0
Hi @vkhorikov,

I have a console application project using your package that I recently upgraded to .NET 6.0. I noticed that when publishing the project **with assembly trimming**, build process generates warnings for the package.

Here's a simple example to reproduce the issue. First the `.csproj` file:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="CSharpFunctionalExtensions" Version="2.25.0" />
  </ItemGroup>

</Project>
```
then a `.cs` file using the `Result<>` class (uses C# 10 syntax):
```csharp
using System;
using CSharpFunctionalExtensions;

var result = GetMessage();
if (result.IsSuccess)
{
    Console.WriteLine(result.Value);
}

static Result<string> GetMessage()
{
    return "Hello World!";
}
```
Then *publishing* it with
```bash
dotnet publish --self-contained --runtime linux-x64 /p:PublishTrimmed=true
```
outputs
```
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /home/dev/dotnet/TrimmingTest/Net6Test/Net6Test.csproj (in 719 ms).
  Net6Test -> /home/dev/dotnet/TrimmingTest/Net6Test/bin/Debug/net6.0/linux-x64/Net6Test.dll
/home/dev/.nuget/packages/csharpfunctionalextensions/2.25.0/lib/net5.0/CSharpFunctionalExtensions.dll : warning IL2104: Assembly 'CSharpFunctionalExtensions' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries [/home/dev/dotnet/TrimmingTest/Net6Test/Net6Test.csproj]
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  Net6Test -> /home/dev/dotnet/TrimmingTest/Net6Test/bin/Debug/net6.0/linux-x64/publish/
```
For me this is problematic mainly because my project is set to treat warnings as errors, so the entire build fails because of this. I didn't investigate if the warning is otherwise justified and can produce faulty output.

I'm executing this with `dotnet` 6.0.100 on Linux. Similar example using `<TargetFramework>net5.0</TargetFramework>` in `.csproj` does not produce the warning, so I suspect that making the package itself target .NET 6.0 and resolving potential upgrade issues will fix it, but I didn't try it.

Thanks in advance for your consideration, I'm a big fan of your work. :)

BR

Caleb9:
Resolves #364 when used in a project targetting .NET 6.0 and using [trimming](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained).

To check that these changes actually solve the problem I created a project referencing the library with `<PublishTrimmed>true</PublishTrimmed>` in `.csproj` file, and used `dotnet publish --self-contained -r linux-x64`. When CSharpFunctionalExtensions has `<IsTrimmable>true</IsTrimmable>` (available in .NET 6+) it doesn't produce trimming warnings (IL2104). Note that to repeat this test I found it necessary to "hard" clean the solution, i.e. remove all `bin` and `obj` directories.

Documentation:
* [Prepare .NET libraries for trimming](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming) (specifically [this](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#set-istrimmable) and [this](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#show-all-warnings) section)

### DIFF
--- a/CSharpFunctionalExtensions.Tests/CSharpFunctionalExtensions.Tests.csproj
+++ b/CSharpFunctionalExtensions.Tests/CSharpFunctionalExtensions.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <LangVersion>8.0</LangVersion>
     <NoWarn>0618</NoWarn>
   </PropertyGroup>
--- a/CSharpFunctionalExtensions/CSharpFunctionalExtensions.csproj
+++ b/CSharpFunctionalExtensions/CSharpFunctionalExtensions.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net40;net45;net461;net472;netstandard2.0;net5.0</TargetFrameworks>
+    <TargetFrameworks>net40;net45;net461;net472;netstandard2.0;net5.0;net6.0</TargetFrameworks>
     <PackageId>CSharpFunctionalExtensions</PackageId>
     <Authors>Vladimir Khorikov</Authors>
     <Description>CSharpFunctionalExtensions - functional extensions for C#</Description>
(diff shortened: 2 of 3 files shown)


hankovich (approved):

vkhorikov:
Thanks! Merging and pushing a new version now.

PR #382, opened 2021-12-22
Add type Result as predicate for .Ensure()

### LINKED ISSUE #337: Add type Result as predicate for .Ensure()
At the moment the `.Ensure()` method signature looks like this:
`public static Result Ensure(this Result result, Func<bool> predicate, string errorMessage)`

It takes a boolean predicate (`Func<bool>`) and an error message string.
It returns a new failure result with the provided error message if the predicate is false. Otherwise, it returns the starting result.

The method should also be able to take a predicate of type Result (`Func<Result>`). If the predicate is a failure result it returns a new failure result with the error message from the predicate, otherwise it returns the starting result.

The method signature would be:
`public static Result Ensure(this Result result, Func<Result> predicate)`

DerStimmler:
This PR resolves #337 by adding the possibility to use a predicate of type `Result` in the `.Ensure()` method.

It's implemented for all sync & async variants of the method and backed up with unit tests.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/EnsureAsyncBothTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/EnsureAsyncBothTests.cs
@@ -1,0 +1,97 @@
+﻿using System.Threading.Tasks;
+using FluentAssertions;
+using Xunit;
+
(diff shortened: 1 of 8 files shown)


vkhorikov:
Thank you! Will publish this with a new version shortly.

PR #384, opened 2022-01-11
Bind-/Map-/WithTransactionScope() fixed for .NET 5 and later (NET5_0_…

mnissl:
…OR_GREATER)


vkhorikov:
Thank you! Will publish this shortly.

PR #385, opened 2022-01-18
WithTransactionScope: isolation level "READ COMMITTED" instead of "SE…

mnissl:
…RIALIZABLE".

For the motivation behind this pull request, please see:

[using new TransactionScope() Considered Harmful](https://docs.microsoft.com/en-us/archive/blogs/dbrowne/using-new-transactionscope-considered-harmful)

My code changes correspond to the fix suggested by "Josh the coder":

[TransactionScope Considered Annoying](https://joshthecoder.com/2020/07/27/transactionscope-considered-annoying.html)

### DIFF
--- a/CSharpFunctionalExtensions/Result/Extensions/WithTransactionScope.cs
+++ b/CSharpFunctionalExtensions/Result/Extensions/WithTransactionScope.cs
@@ -7,10 +7,16 @@ namespace CSharpFunctionalExtensions
 {
     public static partial class ResultExtensions
     {
+        private static readonly TransactionOptions _transactionOptions = new TransactionOptions
+        {
+            IsolationLevel = IsolationLevel.ReadCommitted,
+            Timeout = TransactionManager.DefaultTimeout
+        };
+
         private static T WithTransactionScope<T>(Func<T> f)
             where T : IResult
         {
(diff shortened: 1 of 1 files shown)


hankovich:
Thanks for the articles you mentioned. I have read them carefully and I think they are too biased and too focused on MS SQL Server. 

Do you have any concerns about using the defaults for any other RDBMS?

mnissl:
Well, the isolation level "serializable" hinders concurrency which can lead to performance issues like deadlocks and timeouts.

The default isolation level in SQL Server is "read committed" ... the [default in MySQL](https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-isolation-levels.html) is "repeatable read".

Having TransactionScope default to "serializable" is quite harsh, IMHO.

Maybe we should make it configurable, in a first step globally through a public static variable, as it is the case with `Result.DefaultConfigureAwait`.

vkhorikov:
Sorry for late replies, let me read the articles.

vkhorikov:
Well, you shouldn't really use ambient transactions anyway. But I agree, realizable is not a typical the isolation level anyone would expect. I certainly didn't expect it to be that. Let's change it to read committed, which is the default in most databases.

PR #386, opened 2022-01-20
Make Result.DefaultTryErrorHandler Configurable

YudApps:
Many times when I am using Try/OnSuccessTry I want to have a consistent different error handling than `exc=>exc.Message`.
I started using my own constant `CSharpFunctionalExtensionsConfiguration.ErrorHandler` but sometimes the developers just don't use it by mistake. 
I don't want to have my own XResult.Try(Act) just to have my own ErrorHandler behavior by calling Result.Try(Act, MyHandler).
I think we should allow this configuration to be just part of Result.


vkhorikov:
Looks good, merging. Will be published shortly

PR #400, opened 2022-02-24
Finally(): superfluous async/await removed in static class AsyncResul…

mnissl:
…tExtensionsRightOperand.


mnissl:
My changes are correct if DefaultAwait() performs ConfigureAwait(false).

I'm afraid they are wrong if ConfigureAwait() makes the task return to the calling thread, even if there are no further statements in the Finally() method to execute, but potential chained code after Finally() requires to be executed on the calling thread.

vkhorikov:
Merging, thank you.

PR #402, opened 2022-03-04
Correct broken Execute example in README

staffanwingren:
Minor typo in example of the usage of Execute in README as you cannot call .ToString() on the delegate.


vkhorikov:
Merged, thank you.

PR #403, opened 2022-03-07
adding UnitResult.Ensure()

Yaevh:


vkhorikov:
Could you change `IUnitResult` to `UnitResult`? Interfaces such as `IUnitResult` aren't really meant as a target for extension methods.

Yaevh:
It's actually something I wondered about - why `Result`, `UnitResult` etc. all implement various interfaces (`IResult<T>`, `IValue<T>` etc.), but extension methods do not target them? Why interfaces aren't really meant as a target for extension methods? Is it a general rule, or just a specific rule for `CSharpFunctionalExtensions`?
Or maybe I should just change the code and open a separate discussion thread? :)

vkhorikov:
Those interfaces are mainly for implicit conversion in cases with C# dynamics (e.g https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Result/ResultTE.cs#L41-L47 ).

> Is it a general rule, or just a specific rule for CSharpFunctionalExtensions?

It is a general rule, yes. The idea is that you should return the most specific type possible from your methods, so that consumers of those methods have the richest functionality possible at their disposal.

Extensions *could* target interfaces (and return the specific `UnitResult`) but it's unnecessary since all the library code works with the specific types anyway.

PR #411, opened 2022-05-05
ResultExtensions: Deconstruct() overloads without isFailure

mnissl:
While having the two properties `IsSuccess` and `IsFailure` on the type `Result` eases reading in conditional statements, deconstructing a `Result` object ends up with the two properties being transformed into two variables in the same scope where


vkhorikov:
Was there something wrong with the code?

PR #415, opened 2022-06-10
Avoid multiple enumeration

### LINKED ISSUE #413: Double enumeration in TryFirst
Current implementation of [`TryFirst`](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Maybe/MaybeExtensions.cs#L330-L338) extension method leads unnecessary double enumeration (`source.Any()` and `source.First()`). It has no problem when source is a collection but when I use it in chain with LINQ it leads double applying of LINQ methods.

E.g.
```csharp
var myCollection = new List<int> { 4, 8, 15, 23, 42 };

myCollection
  .Where(i => i % 2 == 0) // Called twice
  .Select(i => i * 10) // Called twice
  .TryFirst();
```

alexmurari:
Fixes #413


vkhorikov:
Merging, thank you. Will be published shortly.

PR #421, opened 2022-06-28
Inline payload retrieval

ProphetLamb:
Allows for branchless inline payload retrieval via the `TryGetValue` API pattern used for [Dictionaries](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue), [JSON Objects](https://docs.microsoft.com/en-us/dotnet/api/system.json.jsonobject.trygetvalue) and many more.

The API vastly simplifies conditional retrieval in a fluent manner, as can be seen here:
```csharp
Maybe<string> MethodMayFail() { [...] }

if (MethodMayFail().TryGetValue(out string success))
{
    Console.WriteLine($"Its alive! And its named {success}");
} else {
   Console.WriteLine("No revival here");
}
```
Previously a temporary, badly named variable was necessary as C# does not support overloading of stack variables.
```csharp
Maybe<string> MethodMayFail() { [...] }

Maybe<string> maybeSuccessMaybeFailure = MethodMayFail();
if (maybeSuccessMaybeFailure.HasValue)
{
    string success = maybeSuccessMaybeFailure.Value;
    Console.WriteLine($"Its alive! And its named {success}");
} else {
   Console.WriteLine("No revival here");
}
```
A few issues related to control flow arise with the previous example.
The `HasValue` call does not syntactically guarantee that `maybeSuccessMaybeFailure` indeed contains a value, therefore a duplicate check is required when obtaining `Value` with the possibility of a thrown exception.
Which in turn would technically need to be handled, or added to the method documentation, thought semantically we know it is never to be thrown in the example above.
This results in separation of control flow and data flow form `HasValue` until `Value` is called.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
@@ -172,6 +172,46 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             
             act.Should().NotThrow();
         }
+        
+        public void TryGetValue_returns_false_if_source_is_empty()
+        {
+            var maybe = Maybe<int>.None;
+
+            bool result = maybe.TryGetValue(out int value);
+            
+            result.Should().BeFalse();
+            value.Should().Be(default);
+        }
+        
+        public void TryGetValue_returns_true_if_source_has_value()
+        {
+            var maybe = Maybe.From(5);
+
+            bool result = maybe.TryGetValue(out int value);
+
+            result.Should().BeTrue();
+            value.Should().Be(5);
+        }
+        
+        public void TryGetValue_returns_false_if_source_is_empty_and_out_parameter_is_null()
+        {
+            var maybe = Maybe<int?>.None;
+
+            bool result = maybe.TryGetValue(out int? value);
+            
+            result.Should().BeFalse();
+            value.Should().BeNull();
+        }
+        
+        public void TryGetValue_returns_true_if_source_has_value_and_out_parameter_is_null()
+        {
+            var maybe = Maybe<int?>.From(5);
+
+            bool result = maybe.TryGetValue(out int? value);
+
+            result.Should().BeTrue();
+            value.Should().Be(5);
+        }
 
         private class MyClass
         {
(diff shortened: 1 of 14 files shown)


ProphetLamb:
Additionally, for the `Result` types a `Match`, `Unbox`, or `Unwrap` function fulfills the same purpose in reducing the effort of handling a function with this signature
```csharp
public Result<IMongoCollection<BsonDocument>> GetCollection(string category)
```
from this
```csharp
 public async Task<IAsyncCursor<BsonDocument>> Filter(string category, BsonDocument filter, CancellationToken ct = default)
{
    var coll = this.GetCollection(category);
    if (coll.IsFailure)
    {
        this.logger.ErrorInvalidCategory(category, coll.Error);
        return EmptyCursor<BsonDocument>.Instance;
    }
    return await coll.Value.FindAsync(filter, DefaultFindOptions,  ct);
}
```
to this 
```csharp
 if (!this.GetCollection(category).Unpack(out var coll, out var error))
{
    this.logger.ErrorInvalidCategory(category, error);
    return Task.CompletedTask;
 }
return coll.InsertManyAsync(documents.Where(this.FilterValidDocuments), DefaultInsertManyOptions, ct);
```

using a simple extension method for all four Result types respectively 
```
public static bool Unpack<T>(in this Result<T> result, [NotNullWhen(true), MaybeNullWhen(false)] out T value,
        [NotNullWhen(false), MaybeNullWhen(true)] out string failure)
{
    if (result.IsFailure)
    {
        failure = result.Error;
        value = default!;
        return false;
    }
    failure = default!;
    value = result.Value!;
    return true;
}
```

vkhorikov:
Looks that there are merge conflicts now. Could you please resolve them?

Regarding `Unpack` -- looks good, but how about calling it `TryGetValue`, to be consistent with `Maybe`?

ProphetLamb:
The conflicts are resolved.

> Regarding `Unpack` -- looks good, but how about calling it `TryGetValue`, to be consistent with `Maybe`?

Contrary to `Maybe` the `Result` monad cannot have a fail condition when extracting data; either the Success or Failure value are always obtained. The wording `TryGetValue` is misleading in this context. But I agree that `Unpack` is unfortunate wording. 

On second thought, a method group `TryGetSuccess` with identities such as 
- `(out TSuccess) -> Boolean` and
-  `(out TSuccess, out TFailure) -> Boolean`

would be sensible, if the negative `TryGetFailure` is also present, with identities such as
- `(out TFailure) -> Boolean` and
-  `(out TFailure, out TSuccess) -> Boolean`.

This way the naming style is consistent, so that IntelliSense can act on the `TryGet` prefix.

@vkhorikov would that be preferable?

vkhorikov:
>would that be preferable?

Yep, both `TryGetValue` and `TryGetError` look good to me.

PR #422, opened 2022-06-28
Fix unessesary shadowing

ProphetLamb:
Mainly declaring structs as `readonly` and all functions as contravariant, so that the compiler does not create shadow copies per call, instead passes the stack pointers on.
For more information on the performance and memory benefits, here is some further reading
[In parameter modifier](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier)
https://devblogs.microsoft.com/premier-developer/the-in-modifier-and-the-readonly-structs-in-c/

During that, I noticed that reflection performed during equality checking. Which is not good at all. Upon further exploring, I noticed, that during the equality comparison from `Maybe` to `object` the value is boxed up to two times, and one reflection object is created, and not cached using the  `Objeckt.GetType` method.
This is circumvented using `EqualityComparer<T>.Default` which is a lazily initialized `IEqualityComparer<T>` that performs a `Buffer.MemCmp` for structs and `Object.ReferenceEquals` for classes that do not implement `IEqualtable<T>`; otherwise it compiles an expression of the `Equals` method and wraps it in a `IEqualityComparer<T>`. Per generic type, we have a reasonable timeout because of the JIT on first call, even that is tiny compared to reflection. The functionallity of `EqualityComparer<T>.Default` can be explored [here](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMAWAbj0JMwDoAlaYASwFsBTOgYQHs2ADiwA2HMAGUxANxYBjDgGcquPAuCRZwIgEE8Aej1EjREEQCSAUQCOEAIbBbAI1EAebQD48AbzzGiLKC0AfSlbYQgOZT9UAGYdAAoArVDwjgBKIi8iELCIogBeIhSIiiIAXzx9Q2NYokdeXmEiazthBXjtIl5gAAsxNKq/H1wDP2iAdmzijgLC7r6wOhzU5VHjCtwN6nQiACFMvC38ONIANhId7gPcaPQd4bHjTqkCojgOADNbCGFgKMfaABOeJmBQAOR+wniUjSaX+j32Uh2hSgkPhY1IwNBEOEUKRsPRfk6UBeKI4AHd4pg4b4AZiQeDIfESQTaQiiCTkRyKfEaTc6ZgsYzccz8Xy/Bs/GyiABtACyHF6vDgZkEUIVSpVaoA8gJWLwoAo6NoAOYmsCKBQsKQcMxQYQBAImtIAXWltTOdQaTWxkJcABV3IkoER/UVculrgCiNKJs0bGEWMAAJ58QS2C1gAPuOgAEU+31+dBaYXa7y+P2AMHDqXF60OQA==)

### DIFF
--- a/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
+++ b/CSharpFunctionalExtensions.Tests/MaybeTests/BasicTests.cs
@@ -161,6 +161,7 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             result.Should().Be(value);
         }
       
+        [Fact]
         public void Maybe_None_doesnt_throw_on_Deconstruct()
         {
             Maybe<int> maybe = Maybe.None;
@@ -173,6 +174,7 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             act.Should().NotThrow();
         }
         
+        [Fact]
         public void Maybe_struct_default_is_none()
         {
             Maybe<int> maybe = default;
@@ -181,6 +183,7 @@ namespace CSharpFunctionalExtensions.Tests.MaybeTests
             maybe.HasNoValue.Should().BeTrue();
         }
         
+        [Fact]
         public void Maybe_struct_value_is_some()
         {
             Maybe<int> maybe = 5;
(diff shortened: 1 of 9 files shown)


vkhorikov:
Sorry for the late reply. Looks good, merging, thank you.

vkhorikov:
I had to revert `public static implicit operator Maybe<T>(T value)` back because the `Bug_MaybeDynamicNone` failed otherwise. Feel free to amend this PR or open a new one if you have further suggestions for that operator.

PR #423, opened 2022-07-06
Rework: Fix unessesary shadowing

ProphetLamb:
This PR fixes the remaining issue from [#422 Fix unessesary shadowing](https://github.com/vkhorikov/CSharpFunctionalExtensions/pull/422) where `implicit T: Maybe<T>` would not accept dynamic typed box of `T` as a parameter when used by the `DynamicBinder` at runtime.


vkhorikov:
Looks good now, thanks!

PR #434, opened 2022-08-16
Add extension methods explicitly converting System.Nullable

ProphetLamb:
Given a extension method accepting the lambda with the identity `T: Maybe<U>` called `FilterMap`, and the expression of the lambda `Expr` with the identity `T: U?`:
```csharp
public static delegate void FilterMap<T, U>(Func<T, Maybe<U>> filter);

IEnumerable<Process> ps = Process.GetProcesses();
ps.FilterMap(static p => p.MainModule?.FileName);
```
`FilterMap` cannot be called with `Expr` because `Expr` doesn't match the identity, to fix this explicit type parameter can be added so that the `implicit operator` for `T: Maybe<T>` is called.

```csharp
ps.FilterMap<Process, string>(static p => p.MainModule?.FileName);
```
Explicit type parameters are undesirable codestyle, when the type is evident.

Explicitly converting the `T` to `Maybe<T>` circumvents the limitation of the language.
This may be done using non-conflicting extension method of the identity `Nullable<T>: Maybe<T> where T: struct` and `T?: Maybe<T> where T: class`.

This PR implements the abovementioned extension methods, so that the following syntax may be used.
```csharp
ps.FilterMap(static p => (p.MainModule?.FileName).ToMaybe())
```
Please note that the extension methods are not available on `struct`s, except when nullable, but are always visible for `class`es.
| Available | Unavailable | 
| --- | --- | 
| `int?` | `int` |
| `Nullable<int>` | `int` |
| `RuntimeTypeHandle?` | `RuntimeTypeHandle` |
| `Process` | --- | 
| `Process?` | --- |


ProphetLamb:
Some equivalent to `FilterMap(IEnumerable<T>, Expr): IEnumerable<U>` should be added accepting lambdas `Expr` of the following identities
1. `T: Maybe<U>`
2. `T: Nullable<U> where U: struct` calling 1. with `Expr.ToMaybe()`
3. `T: U? where U: class` calling 1. with `Expr.ToMaybe()` 

`FilterMap` chooses to `yield` to `IEnumerable<U>` depending on `Expr`.
```csharp
foreach (var x in seq)
{
  if (f(x).TryGetValue(out var v))
  {
    yield return v;
  }
}
```
I am not sure about the naming of `FilterMap`, because we already have `Choose` and `Select`, feedback would be appreciated.

vkhorikov:
Looks good, merging this.

PR #440, opened 2022-08-30
Move ValueTask to own namespace

MarcelRoozekrans:
- Added nameValueTask params to differentiate when both namepsaces are imported.

Please note that the Result.Try ambigious test is impossible to fix. 
My suggestion would be in this case if the valueTask named param is not an option we should remove one of the overloads.


vkhorikov:
Looks good. I've pushed a minor fix (directly to master) and also returned AmbiguityTests back so that they don't compile.

Regarding `Result.Try`, I suggest removing the overload with ``ValueTask``s. We'll refactor it into extension methods sometime later. What do you think?

MarcelRoozekrans:
I agree with you! Problem with the .Try method is that its used statically from a partial class. And we cannot chare the partial class over namespaces this way it will never work. 

Thoughts on the refactor because it uses static invocation I do not think it will work as an extension method :)

vkhorikov:
Yeah, indeed. Not an extension method but a static method on a separate class from `Result`. Something like `Try` as @hankovich suggested.

PR #444, opened 2022-09-04
Fix conditional compliation symbols

Yaevh:


vkhorikov:
Merging, thank you. Will publish the fix today.

PR #446, opened 2022-09-16
MapTry feature

rutkowskit:
Added MapTry Result extensions as a next step to obsolete OnSuccessTry extensions.
Fixed Unit tests issue for global Result configuration Tests  (occasionally some test faulted when running in parallel). Those tests are now forced to run non-parallel.
Fixed BindTry ValueTask function parameter naming to taskValue.


vkhorikov:
Sorry for the late reply. Merging this, will publish it shortly.

PR #451, opened 2022-10-16
Additional Functional Extensions

max-arshinov:
PipeTo and PipeToAsync functions as substitutes for |> operator

### DIFF
--- a/CSharpFunctionalExtensions.Benchmarks/CSharpFunctionalExtensions.Benchmarks.csproj
+++ b/CSharpFunctionalExtensions.Benchmarks/CSharpFunctionalExtensions.Benchmarks.csproj
@@ -1,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="BenchmarkDotNet" Version="0.13.2" />
+    </ItemGroup>
+
+    <ItemGroup>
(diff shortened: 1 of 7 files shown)


hankovich:
Hi, Max! I'm glad to see you here.

I have a few concerns about the added code:
1. I'm not a big fan of putting anything to the `System` namespace.
2. `Pipe` seems to be more idiomatic than `PipeTo`.
3. There are currently no methods in the library whose name ends with `Async`.
4. `EitherOr` doesn't fit well with the naming conventions used here. You could probably use `Match` instead.
5. `EitherOr` overload that checks that `x != null` doesn't seem intuitive for me. It may be named `MatchNotNull` to make the condition clear to callers.

I'm ok with adding `Pipe` and `Match` methods, it may be handy and they are definitely C# functional extensions. However, this library is more focused on the `Maybe`/`Result`/`ValueObject` types and extensions for them, so I'm not 100% sure `CSharpFunctionalExtensions` is the best place for `Pipe` and `Match`.  Let @vkhorikov decide.

Since the added methods extend `T`, and some of the users (including myself) don't seeing polluted intellisence, I suggest moving them not to the `CSharpFunctionalExtensions` namespace, but to a new sub-namespace. It can also help to get rid of possible conflicts between `Match` for results/maybes and for `T`.

max-arshinov:
@hankovich thank you for the feedback. I am happy to make all the adjustments you mentioned, provided that @vkhorikov is ok with this functionality as a part of the library. I think adding the WinstonPuckett.PipeExtensions package just to add the Pipe support is overkill. It makes more sense to just copy/paste this code rather than to introduce yet another dependency. On the other hand, it's perfectly fine to have this code as a part of a more extensive library (like this one). I am a heavy user of all the stuff that contains here. I only added the last missing link (at least it's the case for me).

vkhorikov:
The Pipe extensions look useful, but this library is not a good place for them. Many people already complain that the library has become too bloated (with the `Entity`, `ValueObject` base classes in addition to the more appropriate "functional" `Maybe` and `Result`). I don't want to make it even more bloated by adding extensions for `T` or `object`.

max-arshinov:
Closing this pull request then.

PR #453, opened 2022-10-16
DDD improvements

max-arshinov:
- IEntity is introduced for situations when an existing codebase needs to be refactored. 
- AggregateRoot and IAggregateRoot are essential for domain event handling. TDomainEvent instead of IDomainEvent makes it compatible with other libraries, for example with MediatR (which uses INotification)

### DIFF
--- a/CSharpFunctionalExtensions/AggregateRoot.cs
+++ b/CSharpFunctionalExtensions/AggregateRoot.cs
@@ -2,7 +2,12 @@ using System.Collections.Generic;
 
 namespace CSharpFunctionalExtensions
 {
-    public abstract class AggregateRoot<TDomainEvent> : Entity
+    public interface IAggregateRoot<TDomainEvent>
+    {
+        IReadOnlyList<TDomainEvent> DomainEvents { get; }
+    }
+
+    public abstract class AggregateRoot<TDomainEvent> : Entity, IAggregateRoot<TDomainEvent>
     {
         private readonly List<TDomainEvent> _domainEvents = new List<TDomainEvent>();
         public virtual IReadOnlyList<TDomainEvent> DomainEvents => _domainEvents;
--- a/CSharpFunctionalExtensions/AggregateRoot.cs
+++ b/CSharpFunctionalExtensions/AggregateRoot.cs
@@ -1,0 +1,20 @@
+using System.Collections.Generic;
+
+namespace CSharpFunctionalExtensions
+{
+    public abstract class AggregateRoot<TDomainEvent> : Entity
+    {
+        private readonly List<TDomainEvent> _domainEvents = new List<TDomainEvent>();
+        public virtual IReadOnlyList<TDomainEvent> DomainEvents => _domainEvents;
+
+        protected virtual void Raise(TDomainEvent newEvent)
+        {
+            _domainEvents.Add(newEvent);
+        }
+
+        protected virtual void ClearEvents()
+        {
+            _domainEvents.Clear();
+        }
+    }
+}
--- a/CSharpFunctionalExtensions/Entity/Entity.cs
+++ b/CSharpFunctionalExtensions/Entity/Entity.cs
@@ -1,6 +1,6 @@
 ﻿namespace CSharpFunctionalExtensions
 {
-    public abstract class Entity<TId>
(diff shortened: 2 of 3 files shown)


max-arshinov:
@hankovich you are right that Equals and GetHashCode are methods from the Object class. However, I wanted to emphasize Equals and GetHashCode must be overridden for all classes that implement IEntity. Technically this interface doesn't enforce this because all classes are derived from Object anyway. However, their existence in the interface definition makes this requirement more explicit. Perhaps, adding a summary to methods might help to highlight this requirement ever more.

maxime-poulain:
Merge this will change in the `Entity` class's API.
`IsTransient()` 's visiblity goes from `private` to `public`.
Dunno if it is something we want.

jeffward01:
> Merging this will change in the `Entity` class's API. `IsTransient()` 's visiblity goes from `private` to `public`. Dunno if it is something we want.

Personally, I would not mind that.  I work with EFCore pretty often, and I like to add checks when im persisting data to ensure that the `Id` is not null by using `IsTransient()` -- here is an example of this:

```csharp
/// <summary>
		///     Adds or updates the given item.
		/// </summary>
		/// <typeparam name="TAggregateRoot">The type of the aggregate root.</typeparam>
		/// <typeparam name="TKey">The type of the ID.</typeparam>
		/// <param name="repository">The repository.</param>
		/// <param name="item">The item.</param>
		public static async Task AddOrUpdate<TAggregateRoot, TKey>(this IRepository<TAggregateRoot, TKey> repository, TAggregateRoot item)
			where TAggregateRoot : AggregateRoot<TAggregateRoot, TKey>
			where TKey : IComparable<TKey>, IEquatable<TKey>
		{
			Guard.Against.Null(repository);
			Guard.Against.Null(item);

			if(item.IsTransient)
			{
				await repository.AddAsync(item);
			}
			else
			{
				await repository.UpdateAsync(item);
			}
		}
```

By having the method `IsTransient()` exposed within `CSharpFunctionalExtensions` would encourage greater use and coupling of `CSharpFunctionalExtensions` which I think is a good thing, it allows the user to become more functionally oriented (Such as the method above is not Functional Programming Oriented).

maxime-poulain:
In my own entity base class I ship the `IsTransient() ` method with a public visibility.

For cases like change tracking is disabled helps. 

I suggest however that the method is `virtual`

vkhorikov:
Yeah, we can make `IsTransient()` public and virtual, I don't see any issues with this.

maxime-poulain:
> > Merging this will change in the `Entity` class's API. `IsTransient()` 's visiblity goes from `private` to `public`. Dunno if it is something we want.
> 
> Personally, I would not mind that. I work with EFCore pretty often, and I like to add checks when im persisting data to ensure that the `Id` is not null by using `IsTransient()` -- here is an example of this:
> 
> ```cs
> /// <summary>
> 		///     Adds or updates the given item.
> 		/// </summary>
> 		/// <typeparam name="TAggregateRoot">The type of the aggregate root.</typeparam>
> 		/// <typeparam name="TKey">The type of the ID.</typeparam>
> 		/// <param name="repository">The repository.</param>
> 		/// <param name="item">The item.</param>
> 		public static async Task AddOrUpdate<TAggregateRoot, TKey>(this IRepository<TAggregateRoot, TKey> repository, TAggregateRoot item)
> 			where TAggregateRoot : AggregateRoot<TAggregateRoot, TKey>
> 			where TKey : IComparable<TKey>, IEquatable<TKey>
> 		{
> 			Guard.Against.Null(repository);
> 			Guard.Against.Null(item);
> 
> 			if(item.IsTransient)
> 			{
> 				await repository.AddAsync(item);
> 			}
> 			else
> 			{
> 				await repository.UpdateAsync(item);
> 			}
> 		}
> ```
> 
> By having the method `IsTransient()` exposed within `CSharpFunctionalExtensions` would encourage greater use and coupling of `CSharpFunctionalExtensions` which I think is a good thing, it allows the user to become more functionally oriented (Such as the method above is not Functional Programming Oriented).

Careful if you are using `Guid` for you ID's, for example, using `context.DbSet.Add(SomeEntity);` will generate for you a new ID.

This means that your `IsTransient` will say `false` but your entity has not been yet persisted !

PR #460, opened 2022-10-31
Update README to reflect maintained testing framework

KyleMcMaster:
Updates the README to include a reference to `CSharpFunctionalExtensions.FluentAssertions`, a set of extensions to make test assertions more fluent when using CSharpFunctionalExtensions. This removes a reference to the no longer maintained `FluentAssertions.CSharpFunctionalExtensions`, which works well but has not been updated in quite some time and lacks many of the latest additions like `Result<T, E>`, `Unit Result`, etc. Note that the project name is an h3, so that other testing frameworks could be included as well under the `## Testing` h2 header in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,21 +547,26 @@ Console.WriteLine(bananaInventory.MapError(ErrorEnhancer).ToString()); // "Faile
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


vkhorikov:
Thank you.

PR #464, opened 2022-11-17
Adding optional comparer to Maybe Equals and GetHashcode

julienasp:
I found my self having to write my own Maybe IEqualityComparer for a specific DTO of T. 

But I didn't want to bloat that DTO with equality comparison functionality, so i introduced a IEqualityComparer for that T type. But i couldn't give it to the Maybe.Equals and GetHashCode

### DIFF
--- a/CSharpFunctionalExtensions/Maybe/Maybe.cs
+++ b/CSharpFunctionalExtensions/Maybe/Maybe.cs
@@ -142,7 +142,7 @@ namespace CSharpFunctionalExtensions
             return false;
         }
 
-        public bool Equals(Maybe<T> other)
+        public bool Equals(Maybe<T> other, IEqualityComparer<T> comparer = null)
         {
             if (HasNoValue && other.HasNoValue)
                 return true;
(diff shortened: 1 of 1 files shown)


CSharpFunctionalExtensions/Maybe/Maybe.cs
```diff
@@ -142,23 +142,23 @@ public override bool Equals(object obj)
...
+        public bool Equals(Maybe<T> other, IEqualityComparer<T> comparer = null)
         {
             if (HasNoValue && other.HasNoValue)
                 return true;
 
             if (HasNoValue || other.HasNoValue)
                 return false;
 
-            return EqualityComparer<T>.Default.Equals(_value, other._value);
+            return comparer.Equals(_value, other._value) ?? EqualityComparer<T>.Default.Equals(_value, other._value);
```
hankovich, line 153: `comparer.Equals` throws if `comparer` is null. The same applies for `comparer.GetHashCode`

hankovich:
Could you please provide additional context?

It feels a bit weird to take `IEqualityComparer<T>` to `Equals` and `GetHashCode` methods. The whole idea of `IEqualityComparer` is to separate equality checking from the object itself.

Usually `IEqualityComparer` is accepted by ctors of collections or by methods that do something with collections, but not by instance methods of single objects

PR #465, opened 2022-11-17
Adding optional IEqualityComparer for Maybe

julienasp:
Context:
Given that most Generic<T> class like HashSet<T>, Dictionnary<T>, ThreadSafeDictionary<T>, can be instanciate with something else than the default EqualityComparer.Default. I wanted to add this functionality to the Maybe class.

It adds flexibility, and help to be Open/Close.

Added test:

- Given a T class with object.Equals override then it is used correctly
- Given a T class with object.GetHashCode override then it is used correctly
- Given a Maybe<T> class with a T comparer, then it is used

### DIFF
--- a/CSharpFunctionalExtensions/Maybe/Maybe.cs
+++ b/CSharpFunctionalExtensions/Maybe/Maybe.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 #if NET45_OR_GREATER || NETSTANDARD || NETCORE || NET5_0_OR_GREATER
 using System.Runtime.CompilerServices;
--- a/CSharpFunctionalExtensions/Maybe/Maybe.cs
+++ b/CSharpFunctionalExtensions/Maybe/Maybe.cs
@@ -100,6 +100,11 @@ namespace CSharpFunctionalExtensions
             return new Maybe<T>(obj, equalityComparer);
         }
 
+        public static Maybe<T> From(T obj)
+        {
+            return new Maybe<T>(obj);
+        }
+
         public static bool operator ==(Maybe<T> maybe, T value)
         {
             if (value is Maybe<T>)
--- a/CSharpFunctionalExtensions/Maybe/Maybe.cs
+++ b/CSharpFunctionalExtensions/Maybe/Maybe.cs
@@ -19,8 +19,6 @@ namespace CSharpFunctionalExtensions
 
         private readonly T _value;
 
-        private readonly IEqualityComparer<T> _equalityComparer;
-
         /// <summary>
         /// Returns the inner value if there's one, otherwise throws an InvalidOperationException with <paramref name="errorMessage"/>
         /// </summary>
@@ -68,19 +66,17 @@ namespace CSharpFunctionalExtensions
         public bool HasValue => _isValueSet;
         public bool HasNoValue => !HasValue;
 
-        private Maybe(T value, IEqualityComparer<T> equalityComparer = null)
+        private Maybe(T value)
         {
             if (value == null)
             {
                 _isValueSet = false;
                 _value = default;
-                _equalityComparer = EqualityComparer<T>.Default;
                 return;
             }
 
             _isValueSet = true;
             _value = value;
-            _equalityComparer = equalityComparer ?? EqualityComparer<T>.Default;
         }
 
         public static implicit operator Maybe<T>(T value)
@@ -94,11 +90,6 @@ namespace CSharpFunctionalExtensions
         }
 
         public static implicit operator Maybe<T>(Maybe value) => None;
-
-        public static Maybe<T> From(T obj, IEqualityComparer<T> equalityComparer = null)
-        {
-            return new Maybe<T>(obj, equalityComparer);
-        }
 
         public static Maybe<T> From(T obj)
         {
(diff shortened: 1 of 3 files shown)


julienasp:
@hankovich FYI

CSharpFunctionalExtensions/Maybe/Maybe.cs
```diff
@@ -90,9 +95,9 @@ private Maybe(T value)
 
         public static implicit operator Maybe<T>(Maybe value) => None;
 
-        public static Maybe<T> From(T obj)
+        public static Maybe<T> From(T obj, IEqualityComparer<T> equalityComparer = null)
```
hankovich, line 98: It's not backward compatible. This code will not compile anymore. 

```csharp
Foo(Maybe<int>.From)

void Foo(Func<int, Maybe<int>> selector) 
{
} 
```

Let's add a new overload instead.

hankovich (commented):
BTW the build is broken

julienasp:
> BTW the build is broken

yeah i'm still trying to figure out why, it looks like my changes broke some .NET 5 and above behavior with 

```
#if NET5_0_OR_GREATER
        public static Nullable<T> AsNullable<T>(ref this Maybe<T> value)
            where T : struct
        {
            return System.Runtime.CompilerServices.Unsafe.As<Maybe<T>, Nullable<T>>(ref value);
        }
#else
```

julienasp:
@ProphetLamb  any idea how we could make it work?

hankovich:
In the case when two Maybe's were created with different equality comparers which one should be used to compare them?

ProphetLamb:
The equals operator is defined as non-associative, this implies that two maybes over the same value are equal. 
```csharp
// This one is case-insensitive
Maybe<string> lhs = new("Hello Mom", StringComparison.InvariantIgnoreCase);
// This one is case-sensitive
Maybe<string> rhs = new("hello mom", StringComparison.Invariant);
// Comparing the 1st with the 2nd evaluates true
lhs.Equals(rhs).Should().BeTrue();
// Comparing the 2nd with the 1st on the other hand evaluates false, bc the case-sensitive comparison is used 
rhs.Equals(lhs).Should().BeFalse();
```
As you can see this introduces associatively into a non-associative operator. Therefore any equality/sorting behaviour is undefined! Assuming we expand this to `IComparable`.

A more idiomatic solution is wrapping the (equality) comparison.
```csharp
public sealed record DeferMaybeEqualityComparer<T>(IEqualityComparer<T> inner) : IEqualityComparer<Maybe<T>> {
...
}
```
You can then pass it to the collection.
```csharp
HashSet<Maybe<string>> hs = new(new DeferMaybeEqualityComparer(StringComparison.InvariantIgnoreCase));
```

julienasp:
> The equals operator is defined as commutative, this implies that two maybes over the same value are always equal, regardless of the order of operation.
> 
> But, the proposed solution exhibits the behaviour illustrated below.
> 
> ```cs
> // This one is case-insensitive
> Maybe<string> lhs = new("Hello Mom", StringComparison.InvariantIgnoreCase);
> // This one is case-sensitive
> Maybe<string> rhs = new("hello mom", StringComparison.Invariant);
> // Comparing the 1st with the 2nd evaluates true
> lhs.Equals(rhs).Should().BeTrue();
> // Comparing the 2nd with the 1st on the other hand evaluates false, bc the case-sensitive comparison is used 
> rhs.Equals(lhs).Should().BeFalse();
> ```
> 
> As you can see this equals operation is non-commutative. Therefore any equality/sorting behaviour is undefined! Assuming we expand this to `IComparable`.
> 
> Instead of referencing the comparer in each `Maybe`, a more idiomatic solution is wrapping the comparison in a new comparer, then directly including the new comparer in the collection.
> 
> ```cs
> public sealed record DeferMaybeEqualityComparer<T>(IEqualityComparer<T> inner) : IEqualityComparer<Maybe<T>> {
> ...
> }
> ```
> 
> You can then pass it to the collection.
> 
> ```cs
> HashSet<Maybe<string>> hs = new(new DeferMaybeEqualityComparer(StringComparison.InvariantIgnoreCase));
> ```

this is how I work around the limitation @ work, I introduce a comparer for IEqualityComparer<Maybe<T>> that was using in his inner code a T comparer as well.

You both @ProphetLamb  and @hankovich bring some fair concerns about the non-commutative behaviour of my changes. I considered adding in the Maybe definition of equals the comparaison of the IEqualityComparer to make sure that they were both equals otherwise it would return false. But i did not implemented it. Anyway i dont see how i could make it all work considering that adding a field in Maybe.cs break the AsNullable() extension method.

hankovich:
The idea here is that no changes to `Maybe<T>` are required. The single thing that may be added to the library is a `MaybeEqualityComparer<T>` that accepts `IEqualityComparer<T>` as a ctor parameter.

This will cover all your cases.

julienasp:
> The idea here is that no changes to `Maybe<T>` are required. The single thing that may be added to the library is a `MaybeEqualityComparer<T>` that accepts `IEqualityComparer<T>` as a ctor parameter.
> 
> This will cover all your cases.

Indeed, ill implement it today. Thanks for the feedback guys!

ProphetLamb (approved):
Looks good to me

vkhorikov:
Looks good to me too, but let's wait for @hankovich to give a green light as well.

hankovich:
Implementation looks good for me.

I'm not a big fan of tests which
1. Check a lof of cases in a single test
2. Do

```csharp
var result = /* an expression under test */
result.Should().BeTrue();
```

Instead of

```csharp
subject.Should().BeEquivalentTo(expectation, opt => opt.Using(equalityComparer);
```

Documentation clearly [says](https://fluentassertions.com/tips/#general-tips) that.

But I'm ok to merge this as is.

vkhorikov:
Merging, will be published shortly.

PR #488, opened 2023-03-08
Add GetValueOrDefault to Result

michalsznajder:
Based on Maybe. Tests also copied from Maybe.


michalsznajder:
Test are wrong. I will update them.

PR #489, opened 2023-03-08
Add GetValueOrDefault to Result

michalsznajder:
Based on Maybe. Tests also copied from Maybe.


vkhorikov:
Thank you. Will publish shortly with v2.38.0

PR #490, opened 2023-03-08
Fix symmetry violation of ValueObject comparison

teheran:
Comparing two incomparable objects in this way can lead to unpredictable results.

        private int CompareComponents(object object1, object object2)
        {
            ...
            **return object1.Equals(object2) ? 0 : -1;**
        }

Expected if (object1 > object2) then (object2 < object1)

To prevent comparison of non-comparable components of VO it is proposed to change signature of ValueObject.GetEqualityComponents abstract method:

--protected abstract IEnumerable\<object\> GetEqualityComponents();
++protected abstract IEnumerable\<IComparable\> GetEqualityComponents();


vkhorikov:
Thanks much. Will publish this shortly.

PR #492, opened 2023-03-09
add a method overload for GetValueOrThrow, that enables a custom exce…

julienasp:
I added a method overload for GetValueOrThrow, that enables a custom exception instead of the default InvalidOperationException.

**Justifications:**
I found myself needing that feature, in my production code. The leaner the better for me.

since it was not available I had to do

```
if (result.HasNoValue)
{
     throw new DomainException
}

this.VeryNotComplicatedLogic(result.Value)
```
vs
```
this.VeryNotComplicatedLogic(result.GetValueOrThrow(new DomainException))
```


RaphCayou (approved):
LGTM

vkhorikov:
Looks good, merging.

PR #494, opened 2023-03-13
Add ToResult extension to Nullable.

xavierjohn:
With the support of nullable added to C#, it is more natural to use nullable over Maybe extension.
The PR will add support to convert null/nullable to Result.

Removed old .net framework from build target. .Net users can use .net standard library.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/NullableExtensionTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/NullableExtensionTests.cs
@@ -1,0 +1,198 @@
+﻿#if NETCOREAPP3_0_OR_GREATER
+namespace CSharpFunctionalExtensions.Tests.ResultTests.Extensions
+{
+    using System;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Xunit;
+
+    public class NullableExtensionTests
+    {
+        [Fact]
+        public void Convert_nullable_struct_to_result_pass()
+        {
+            // Arrange
+            DateTime? date = DateTime.Now;
+
+            // Act
+            var result = date.ToResult("Date not set.");
+
+            // Assert
+            result.IsSuccess.Should().BeTrue();
+            result.Value.Should().Be(date.Value);
+        }
+
+        [Fact]
+        public void Convert_nullable_struct_to_result_fail()
+        {
+            // Arrange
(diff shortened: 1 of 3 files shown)


vkhorikov:
The old .NET Framework versions are needed for people who can't yet transition to .net core/standard. For example .NET Framework 4.0 doesn't support .net standard at all, while .NET 4.5 only supports .net standard <= 1.1. Please refer to https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#select-net-standard-version

Please revert the changes in `CSharpFunctionalExtensions.csproj`. The PR looks good otherwise.

xavierjohn:
I added 4.0 and 4.5 back. 
net461 & net472 is covered by .net standard 2.

vkhorikov:
I believe the only one that can be removed is `net472`. For .NET Framework 4.6.1, there's this footnote in the document:

> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

As for .net 5.0, while the project will work without targeting .net 5.0 explicitly, the library currently uses .net5-specific functionality to enable some optimizations (search for `NET5_0_OR_GREATER` for examples). This functionality won't be available if there's no explicit .net 5 build.

vkhorikov:
Thank you. Will publish shortly.

xavierjohn:
@vkhorikov  Why are we supporting .net frameworks like .net 4 that is past EOL?
https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework

Keeping old frameworks makes it harder to contribute to the package since people like me don't have any of those old frameworks installed.

xavierjohn:
I remove the old frameworks, build, test and then revert the csproj to bring back the old frameworks and let the build system take care of building.
I do have the option of installing developer packs for older frameworks but why add more stuff that I don't need to my computer.

vkhorikov:
Got it. Maybe my machine have them, hence I didn't notice. I'll start working on v3.0 shortly, I suggest removing the older frameworks in that version. Let's keep the older frameworks in v2.x.

PR #502, opened 2023-05-03
Allow returning a value from Match for UnitResult

piotr121993:
This is a use case this PR fixes:
image


vkhorikov:
Merging, thank you. Will be published soon.

PR #506, opened 2023-06-06
Adding Task support to TapTry and TapIfTry

tinytownsoftware:
I completely missed these when I created this set of extension methods.


vkhorikov:
Thanks, will be published with v2.39.2 shortly.

Issue #509, opened 2023-06-24
Making ValueObject.GetEqualityComponents return IComparable is a breaking change


robertlarkins:
Previously `GetEqualityComponents()` returned `IEnumerable<object>`. Now it returns `IEnumerable<IComparable>`. If the ValueObject holds an Entity and it is used as an Equality Component it causes a compile time error if the Entity doesn't implement IComparable.

What is the appropriate fix here? Every type that ValueObject uses as an Equality Component must now implement IComparable? Could this extend to the CSharpFunctionalExtensions Entity type so entities have it is available out-of-the-box?

vkhorikov:
> What is the appropriate fix here? Every type that ValueObject uses as an Equality Component must now implement IComparable?

Yes, unfortunately, that's the only way. Otherwise, the comparisons can be inconsistent.

> Could this extend to the CSharpFunctionalExtensions Entity type so entities have it is available out-of-the-box?

You mean make `Entity` implement IComparable?

robertlarkins:
> You mean make Entity implement IComparable?

Yes, then an `Entity` could be used in a `ValueObject` as an equality component without needing to explicitly implement `IComparable` on the derived entity type. `Id` would be the property used for the comparison.

Or is there issues I'm not seeing with `Entity` implementing `ICompariable`?

vkhorikov:
Yes, that's a good idea. No particular reason why it's not done yet. Appreciate if you could submit a PR.

vkhorikov:
Thank you. I'll double check everything this weekend and release a new minor version then.

Szuuuken:
Hi,
this is also a breaking change for any kind of list or array ValueObjects.
The following Examples aren't possible any more:

##### Value Object as an byte array
```c#
public class FileContent : SimpleValueObject<byte[]>
{
    public FileContent(byte[] value) : base(value)
    {
    }
}

``` 

##### Value Object as a List of ValueObjects
```c#
public class Thing: SimpleValueObject<string>
{
    public Thing(string value) : base(value)
    {
    }
}

public class MultpleThings: SimpleValueObject<List<Thing>>
{
    public MultpleThings(List<Thing> value) : base(value)
    {
    }
}

``` 

Do you have any suggestions how to solve such scenarios?

vkhorikov:
You'll need to fall back to the base `ValueObject`:

```c#
public class FileContent : ValueObject
{
[...]

PR #510, opened 2023-06-28
Make Entity implement IComparable

robertlarkins:
Having Entity implement IComparable allows entities to be implicitly used as equality components in ValueObjects.

This PR relates to Issue #509.


robertlarkins:
Would be good to have the following checked:
- `Entity.CompareTo` has the desired implementation
- `Entity` generic type `TId` should implement `IComparable<TId>`
- The tests are adequate and their names are appropriate

vkhorikov:
Looks great, merging, will be published shortly with v2.40.0.

PR #520, opened 2023-10-10
Add two additional TapIf.Task functions

NilsVreman:
This PR adds two additional `TapIf.Task` functional extensions to the Result type.

Sorry, if the formatting doesn't entirely match the remainder of the repo, it's my first PR here :)
I tried to keep it as consistent as possible with most other functions I could find under the same extension directory.


vkhorikov:
The formatting looks good. Merging, will publish shortly, thank you!

PR #522, opened 2023-10-28
Switch to license expression for package license

Greybird:
In order to ease retrieval of license by tooling aiming to check them, as well as ensuring that a change to the master branch license would not be retroactively applied to older packages, I would propose to use [PackageLicenseExpression](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file) to pack the nuget package, rather than rely on a reference to the LICENSE file in the master branch of the repository.


vkhorikov:
Sure why not. I don't really intend to ever make this library non-MIT, but we can as well embed this implicit promise explicitly.

Issue #535, opened 2024-03-28
[Proposal] BindZip to store both income and outcome of Bind


hankovich:
I have a lot of code like

```csharp
GetEntity(entityId)
    .Bind(entity => GetNestedEntity(entity.NestedEntityId).Map(nestedEntities => (entity, nestedEntities));

Result<Entity> GetEntity(EntityId id) => throw null!;

Result<NestedEntity> GetNestedEntity(NestedEntityId id) => throw null!;
```

i.e. in further processing I need both income and outcome of `Bind`. I suggest to add a new method `BindZip`

```c#
GetEntity(entityId)
    .BindZip(entity => GetNestedEntities(entityId)); // returns Result<(Entity First, NestedEntity Second)>
```

hankovich:
It's not only me who needs such overload: https://github.com/vkhorikov/CSharpFunctionalExtensions/discussions/534

vkhorikov:
Sorry for the late reply.
Sure, let's add this.

DerStimmler:
I think this issue can be closed since #540 has been merged?

PR #541, opened 2024-05-02
Add Select/SelectMany Result methods to perform do-notation like operation via LINQ

PNZeml:
This one address same problem from a proposal in #535.

```csharp
var result = await (
    from one in Result.Success(1)
    from two in Result.Success(2)
    from three in Task.FromResult(Result.Success(3))
    select one + two + three
);

Assert(result.Value, 6);
```


PNZeml:
Sorry, didn't see this feature already implemented.

Issue #27, opened 2017-04-12
Strong Name


ResiEvil:
Is it possible to strong name the assemblies?

Many Thanks

vkhorikov:
I would prefer not to string name the package. Could this help it your situation: https://github.com/dsplaisted/strongnamer ?

Ironbell:
Worked like a charm and solved the problem for me. Thank you very much!

yakimovim:
What is the reason why you don't want to sign the assembly? It is not very convenient to use two NuGet packages (yours and StrongNamer) instead of one.

vkhorikov:
It's mostly a relic from the older MS days: https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md#1-microsoft-strong-names-their-assemblies-should-i

vkhorikov:
I plan to release a signed version of the library, similar to how Dapper does it. But I don't have an ETA yet unfortunately.

adcorduneanu:
Vladimir, I don't find your reasoning for non-signing strong enough, but more like a personal flavor on your side. I would prefer working with a strong name assembly every day as this gives some assurance that nobody tampered with the existing implementation, as it will require the original key to sign it. In most companies, there are policies in place that will require this one, together with a digital signature (signtool.exe).

The community should have priority in detriment of flavors. 

Having the key in is actually a minimal effort, compared with the possibility of searching for a drop-in replacement for any unsigned library, unbundling the existing package and signing it internally, or using some kind of unsecure NuGet that will do the signature for you.

I would like to see it in the very next release if possible.

vkhorikov:
Well, I did change my opinion, see my last comment.

bothzoli:
Hey @vkhorikov,
I have created a PR (#543) to address this issue.
Please let me know if you're happy with this solution.
I'm not sure if anything is needed for nuget.org to accept a new package with the same prefix.

vkhorikov:
Fantastic work, thanks @bothzoli . This is probably the longest standing unresolved request for this library.

PR #543, opened 2024-05-23
Create strong named assembiles

bothzoli:
# Create strong named NuGet package

To create a strong named NuGet package, a new project `CSharpFunctionalExtensions.StrongName` was added to the solution.
The project references source files from `CSharpFunctionalExtensions.csproj` and it uses the generated signing key (`CSharpFunctionalExtensions.snk`) to sign the assembly.

The `Dockerfile` was updated so that a strong named NuGet package is also built.
The `azure-pipelines.yml` was also modified so that the newly created package is uploaded as both a GitHub release asset and to nuget.org as well.

Minor fixes were applied to the `azure-pipelines.yml` file:
- the authorization header should have `Bearer` instead of token based on the [documentation](https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#upload-a-release-asset)
- the `cURL` upload should use the `--data-binary` switch (note that nuget packages were not added to previous releases either)
- run conditions were added to the publish steps so that the pipeline won't fail if the credentials are not provided, but simply skip the publish steps

The version file `version.txt` was also updated with a patch version bump.

Fixes: #27


hankovich:
New csproj is just a copy of the existing one. I suggest you to move common properties to Directory.Build.props

bothzoli:
Oh yeah, thanks mate. I knew something was amiss.
I'll update the PR shortly.

bothzoli:
I wonder though, wouldn't a `Directory.Build.props` be imported into all projects including `Examples`, `Tests`?
I'm not sure that'd be correct.
Shall I separate the _main_ projects into an `src` folder?
Or use a different `.props` file and explicitly import where needed?

bothzoli:
I've gone with a `Common.Build.props` file at the root of the solution.
Please let me know if you'd prefer a different approach.

bothzoli:
Hey, any thoughts on the PR?
Thanks

vkhorikov:
Hey, sorry, really short on time these days. I'll test it during the upcoming weekend and will push it "to prod" if everything looks good (which it does on the first glance).

vkhorikov:
The current approach looks good, no need to include the common proj into examples and tests, those aren't getting published anyway. Locally everything looks good, merged it to master, let's see how the build goes.

vkhorikov:
The pipeline has published the new nuget package successfully.

bothzoli:
Wow, that's awesome news. I'll start using it right away.
Thank you very much 😊
I'll try to keep my eyes open, but if there are issues opened for the strong named package (that are working correctly with the _regular one_) feel free to tag me in the issue.

PR #555, opened 2024-07-24
Add IError<string> to the Result struct

### LINKED ISSUE #546: Add IError<string> to the Result struct
Why is the Result.cs struct not decorated with the `IError<string>` interface? The `Result<T>` struct does have the `IError<string>` interface decorated.

Both the `Result` and the `Result<T>` structs have the `string Error` property.

We could just add the interface without any issue, like below:
```csharp
[Serializable]
public readonly partial struct Result : IResult, ISerializable, IError<string>
{
```

bothzoli:
Adding `IError<string>` interface to the `Result` struct.

Fixes #546


nsoethofcovadis (approved):

vkhorikov:
Looks good, mering.

PR #565, opened 2024-09-10
Add allocation-free Maybe.Bind overload

marcinjahn:
For now I've added just the Maybe.Bind overload in order for the approach to be validated.
1. Is the naming OK? (`context`)?
2. Is it OK that the logic is duplicated in allocating and non-allocating overloads? Is it possible to not duplicate the code?
3. Are tests alright?

When the approach gets approved, I will be able to create overloads for other extensions.

This PR is the result of https://github.com/vkhorikov/CSharpFunctionalExtensions/discussions/562.


vkhorikov:
Sorry for the later reply. Some extra empty lines aside, everything looks good to me. Merging.

Issue #566, opened 2024-09-17
Question: How to use linq query syntax for Result types


christophercalm:
Is the Select method not supported for linq query syntax for result?

As an example here is code that works with `Maybe`

```csharp
Maybe<string> name = "John";

var upper = from x in name select x.ToUpper();`
```

But does not work with Result

This is the error: 
CS1936	Could not find an implementation of the query pattern for source type 'Result<string>'.  'Select' not found.	

I see the that SelectMany operation is supported but it looks like there is no select. Is this something that you would be open to me adding?

It would help with workflows like this

```csharp
Result<string> GetGreetingMessage(string userName)
{
    var maybeGreeting =
        from name in ValidateName(userName)   // Bind the validated name
        select CreateGreetingMessage(name);   // Map to a greeting message

    // Convert Maybe to Result
    return maybeGreeting.ToResult("Invalid name provided");
}

Maybe<string> ValidateName(string name)
{
    // Check if the name is valid, return Maybe.None if invalid
    return string.IsNullOrWhiteSpace(name) ? Maybe<string>.None : Maybe<string>.From(name);
}
```
but for results.

vkhorikov:
We're trying to keep naming consistent in this library. In the earlier days there were discussions on which naming approach to choose: the C# one (`Select`, `SelectMany`, etc) or the F# one (`Map`, 'Bind', etc). We decided to follow the F# naming convention because it's more consistent (e.g `SelectMany` doesn't really make sense for `Result` and even for `Maybe`).

So all `Select` extensions remaining should really be marked as obsolete and redirected to `Map`.

The downside to this decision is that Linq expressions (`from ... in ... select`) won't work anymore, but people rarely use those. And I suggest to just use regular linq instead.

christophercalm:
The reason for using linq expressions is that you can combine results from many different monads. You can’t do that with the regular linq syntax without multiple select many expressions. I would ask that you please reconsider as the rest of the api is complete in regards to the overloads for linq query expressions.

vkhorikov:
OK, makes sense. Feel free to submit a PR. A couple of implementation notes:

1. `Select` should call `Map` internally; `SelectMany` -- `Bind` (just so that we don't have duplicate implementations)
2. Please leave a note as to why we have `Select` and `SelectMany` -- to support linq expressions

christophercalm:
> OK, makes sense. Feel free to submit a PR. A couple of implementation notes:
> 
[...]

PR #567, opened 2024-09-20
added select overload with projection for result

christophercalm:
This extension allows you to use the linq query syntax for result as we are currently already able to do using maybe. 

Whereas this code worked before
```csharp
Maybe<string> name = "John";

var upper = from x in name select x.ToUpper();
```

This very similar code would not
```csharp
Result<string> name = "John";

var upper = from x in name select x.ToUpper();
```

Adding an overload to Select with a projection allow for greater flexibility in monadic operations using the linq query syntax which enables the user to pass in values from previous operations down the monadic chain. 

This is related to issue [566](https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/566)

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/SelectTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/SelectTests.cs
@@ -1,0 +1,51 @@
+﻿using FluentAssertions;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


vkhorikov:
Answered here: https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/566#issuecomment-2380594685

PR #576, opened 2024-11-08
feat: add lazy overload of Maybe.ToResult

marcinjahn:
Adds `Maybe.ToResult` variants that accept lambdas to create custom errors. It might be beneficial in cases where constructing an error is expensive for any reason.

I see that CRLF -> LF conversion got applied automatically due to `.gitattributes`. Shouldn't the whole repo get converted to LF via some commit to have better diffs?


CSharpFunctionalExtensions.Tests/MaybeTests/Extensions/ToResultTests.Task.cs
```diff
@@ -1,53 +1,75 @@
...
-            result.IsSuccess.Should().BeFalse();
-            result.Error.Should().Be(E.Value);
-        }
-
-        [Fact]
-        public async Task ToResult_Task_custom_failure_returns_success_if_has_value()
-        {
-            var maybe = Maybe<T>.From(T.Value);
-
-            var result = await maybe.AsTask().ToResult(E.Value.AsTask());
```
marcinjahn, line 47: Please verify if the `E.Value.AsTask()` thing was expected. I found it odd, and changed it to just `E.Value`. I did similar change in other test files.
If the `AsTask()` conversion was expected, do let me know, I'll revert this part.

vkhorikov: Not really sure. Maybe a copy-paste from other tests where we test both task- and non-task overloads of the methods. Here, we don't have an overload that accepts a task as the second argument, so we can remove these tests:
- ToResult_Task_returns_success_if_has_value
- ToResult_Task_custom_failure_returns_success_if_has_value

marcinjahn: Alright, I noticed that for Task and ValueTask, there was one redundant test file for each of them - the "Left" file had exactly the same tests as the "non-left" one. As you said, `ToResult` does not have `(Value)Task` as a second argument.

I removed the redundant file.

I fixed `ToResult_Task_returns_success_if_has_value` since it is a valid test
`ToResult_Task_custom_failure_returns_success_if_has_value` was alright I think, left it as it is.

marcinjahn:
> The CRLF -> LF conversion is a mistake, could you please disable it and redo your PR? We may do this conversion, but that should be a separate PR with no code changes so that we can track whitespace and non-whitespace-related changes separately. Definitely let's not do both at the same time

I restored the line endings. Btw, turns out my original changes added CRLF, while the repo already had LF endings. My bad!

vkhorikov:
Thanks, merging this. Will review the other PRs shortly.

PR #578, opened 2024-11-10
feat: add allocation-free overloads of Maybe.Map

marcinjahn:


vkhorikov:
Here too please, no CRLF -> LF conversion

marcinjahn:
> Here too please, no CRLF -> LF conversion

Done

PR #582, opened 2024-11-10
feat: add allocation-free overloads for Result.Map and Result.MapIf

marcinjahn:
Also fixes some test files, like `MapIfTests.Task.Left.cs` where the tested methods where actually the `MapIf.Task.Right` extensions.


vkhorikov:
CRLF -> LF conversion please

marcinjahn:
> CRLF -> LF conversion please

done

PR #583, opened 2024-11-10
feat: add allocation-free overloads for Result.MapError

marcinjahn:


vkhorikov:
CRLF -> LF conversion please

marcinjahn:
> CRLF -> LF conversion please

done

Issue #580, opened 2024-11-10
Tests failures are ignored in CI pipeline


marcinjahn:
At the moment, tests are not even executed due to .NET version mismatch. Tests project targets .NET 6, while only 8.x SDK is available.

> Testhost process for source(s) '/app/CSharpFunctionalExtensions.Tests/bin/Release/net6.0/CSharpFunctionalExtensions.Tests.dll' exited with error: You must install or update .NET to run this application.

Any PR to this repository has misleading build result - it's green even though tests did not run at all.

vkhorikov:
Ooff. This went unnoticed after removing the old .net references

bothzoli:
Oops, that's my mistake, I honestly didn't notice that the tests were failing.
I'll start working on a fix 😉
Sorry about that 😇

bothzoli:
@vkhorikov please see the fix for this in #586.

PR #586, opened 2024-12-06
Fail pipeline if test results file is missing

bothzoli:
The changes in this PR will make sure that the build pipeline fails if no tests were run for any reason.
This will prevent issues like #580 in the future.

The PR also fixes the failing tests project.
This also revealed another issue regarding obsolete `BinaryFormatter`.
New issue #587 was opened to address this.

Resolves: #580


vkhorikov:
Thank you!

Issue #571, opened 2024-09-30
Proposal: Ensure `null` with `Maybe`


DerStimmler:
Currently, the `ToResult` method on `Maybe` returns a success result if the `Maybe` has a value, and a failure result if there is no value.

But what if it should be the other way around?

---

Lets say I want to create a report, but only if no report already exists.

Right now I would probably do something like this:

```csharp
await Result
      .Success()
      .Map(() => service.FindAsync(id)) //returns Report?
      .Ensure(report => report is null, "There is already a report available.")
      .Tap(_ => service.CreateReport())
      ...
```

But this comes with some overhead because I have to start the chain with an empty success result to then map the nullable value and ensure its `null` using the `Ensure` method.

I think it would be more appropriate and compact to use `Maybe` and its `ToResult` method since I'm dealing with a nullable.

But to keep the chain running if the `Maybe` has no value, I basically need an inverted `ToResult` method that returns a success result if there is no value. E.g:

```csharp
await Maybe
      .From(service.FindAsync(id))
      .ToInvertedResult("There is already a report available.") //returns success result if Maybe has no value, otherwise returns failure result with error message
      .Tap(() => service.CreateReport())
      ...
```

---

What do you think about this addition?

`ToInvertedResult` is just an example name because I can't think of anything better. 😄

vkhorikov:
Yeah, sounds useful enough. Feel free to submit a PR. Can't think of a better name than `ToInvertedResult` either :)

bothzoli:
Hey @DerStimmler, @vkhorikov,
I'm happy to give this one a shot in the coming days.

bothzoli:
If I'm correct with my thinking, this will return a vanilla `Result` for `Maybe<T>` and a `UnitResult<E>` when using a custom failure.
Because I cannot populate a `T` from a failed `Maybe<T>` for it to be `Result<T>` (or populate a `T` for a `Result<T, E>` from a `Maybe<T,E>`).

But I believe it would probably make sense to add overrides which could do just that, like:
`Result<T> ToInvertedResult<T>(in this Maybe<T> maybe, string errorMessage, Func<T> compensate)`
or 
`Result<T> ToInvertedResult<T>(in this Maybe<T> maybe, string errorMessage, Func<string, T> compensate)`

Any thoughts?

DerStimmler:
Hi @bothzoli,
thanks for your help.

I think you are right with `Result` and `UnitResult<E>` as return types.

[...]

PR #588, opened 2024-12-12
Create `ToInvertedResult` extension methods on Maybe

bothzoli:
Added `ToInvertedResult` methods and corresponding tests to `Maybe`.
As a housekeeping effort, removed some potentially superfluous test files.

`ToInvertedResult` method returns a `Success` result from a `None` value and a `Failure` result from a `Some` value.

Resolves: #571


bothzoli:
So I based the tests on the tests for `ToResult` and tried to keep them consistent.
That's when I realised that some of the `ToResult` tests (with `Left` in the name) were most likely some copy-paste because they didn't execute other code paths than the other tests.
As a housekeeping commit I thought I'd clean them up to be consistent with the new `ToInvertedResult`.
I'm happy to undo that commit if needed, but I think it just removes code that shouldn't have been there in the first place.

DerStimmler:
Ah, I see. Then that should be fine.
Following the scout rule here 😄

vkhorikov:
@bothzoli Looks like there are some merge conflicts, could you take a look?

bothzoli:
Sure, I'll take a look once I recover from Christmas food coma 😇

bothzoli:
@vkhorikov I fixed the merge conflict, but also since the target branch has changed a bit with #576 I extended `ToInvertedResult` to also have the lazy error overloads.
I found that the unit tests in #576 did not assert the calling of the lazy error function, so I unified all those tests to do that.
I also found that the `ToUnitResult` extensions were not overloaded with the lazy error creation, so I added it there too.

This PR became a bit more complicated than I imagined it to be, but I think overall the code base became more consistent.

vkhorikov:
Thank you!

PR #589, opened 2024-12-28
Fix GitHub asset upload on releases

bothzoli:
Updated the corresponding `cURL` request based on [docs](https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#upload-a-release-asset).
This should hopefully fix the missing asset uploads on releases.


vkhorikov:
Yeah, let's see if this helps. Thanks for looking into this.

bothzoli:
Looks like it didn't work, but now I know what the issue is 🥳
See #590, I'll fix this once and for all soon 😉

vkhorikov:
Looks like it worked this time 🎉

Issue #587, opened 2024-12-06
Remove obsolete BinaryFormatter from tests


bothzoli:
Since .NET 7 `Serialize` and `Deserialize` on `BinaryFormatter` produce compile errors (see [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/7.0/binaryformatter-apis-produce-errors)).

These methods are currently used in the test classes `CSharpFunctionalExtensions.Tests/MaybeTests/DeserializationTests.cs` and `CSharpFunctionalExtensions.Tests/ResultTests/DeserializationTests.cs`.

With #586 this error is switched off to allow tests to succeed, but the corresponding tests should be removed or updated to use a different serializer.

The switching off happens in `CSharpFunctionalExtensions.Tests/CSharpFunctionalExtensions.Tests.csproj` with `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>` added.

bothzoli:
As also mentioned in #584 since `Result` (and also `Maybe`) types are not serializable, I wonder if it makes sense to keep these tests?

vkhorikov:
We can remove them. We shouldn't have made `Result` serializable either btw.

bothzoli:
Thanks, I'm happy to create a PR removing the tests.
Should I also remove the `ISerializable` interface from `Result` or would you prefer to do that for a major release?

vkhorikov:
Please do. Let's keep `ISerializable` for now as it's going to be a breaking change.

bothzoli:
@vkhorikov PR #592 is ready for review.

PR #592, opened 2024-12-29
Remove obsolete deserialization tests

bothzoli:
Serialization and deserialization is based on the obsolete `BinaryFormatter`.
Since `Maybe` and `Result` should not be serialized anyways these tests can be removed along with the obsolete `BinaryFormatter`.

Fixes: #587


vkhorikov:
Looks great, merging

Issue #594, opened 2025-01-10
ValueOr and AsMaybe for results


RorySan:
Hello, I wrote two simple extension methods that are helping me in my project, but I'm guessing there is probably a better alternative already built in that I am not using:

```csharp
    public static Maybe<T> AsMaybe<T>(this Result<T> result) =>
        result.IsSuccess ? result.Value : Maybe<T>.None;
    
    public static string ValueOr<T>(this Maybe<T> maybe, string alternativeValue) 
        where T : SimpleValueObject<string> => 
        maybe.HasValue ? maybe.Value : alternativeValue;
```

I found a way of using the existing Or to achieve this, but it is a bit more verbose. With this ValueOr i get to write things like `Email.ValueOr(string.Empty)` which is handy when displaying optional info on the UI. Am I missing a better way of doing this?

bothzoli:
Hey @RorySan,

There's a `GetValueOrDefault` method defined on `Maybe` that you could use instead of your `ValueOr`.

```cs
public T GetValueOrDefault(T defaultValue)
{
    if (HasNoValue)
        return defaultValue;

    return _value;
}
```

For the `AsMaybe` I did not find anything that would do the same thing, but I may be wrong.
It sounds like a useful enough method.
@vkhorikov or @hankovich do you if there's such a conversion from `Result` to `Maybe` and if not, why not?
Would it make sense to include that?

RorySan:
Thanks for your reply @bothzoli , although I think GetValueOrDefault  wouldn't work in this case because I am not returning T but string. I'll get back to you with the detailed use case and we can decide if it might make sense to include it.

I've been exploring functional programming for a while now and I use this library extensively. I'm not very experienced with contributions, but I'd be happy to help if that's ok.

bothzoli:
In the extension method you sent above, the generic type constraint makes `T` a `SimpleValueObject<string>`, which in turn makes the `T` of the `public T Value` of the value inside the `Maybe` a `string`.
This sounds a bit confusing, but the point is, I believe you could as well use a `Maybe<string>` instead of the `Maybe<SimpleValueObject<string>>`.
In that case the `T` returned by `GetValueOrDefault` will actually be a `string`, as you need it to be.

This all however makes me wonder what you use the `ValueOr` method for? This method would "unwrap" the `Maybe`. Why not just `Map` over it?

vkhorikov:
[...]

PR #597, opened 2025-01-17
ResultAsMaybe

RorySan:
This is a pretty simple implementation. I've explored implementations of other features of the repo and I'd need clarification regarding 3 points:

1. Task and ValueTask implementations, should we have them? I will check similar features in depth and try to replicate them, although I'd appreciate some indications.

2. I've seen `#if NET5_0_OR_GREATER` in a few places, I'm not sure if this is something I should take into account and if or how it affects this feature.

3. Result with both Success and null value: Apparently this is possible, which I guess it makes sense if you want to imply that an operation was successful even though the result is null. As far as I understand (in a Maybe), if the Value is `null`, HasValue has to be `false`. I added a test for this but it might not make sense since it is not possible to create a Maybe containing null.

I know there's a lot missing here. Please point me in the right direction and I'll keep working on it.

### DIFF
--- a/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/AsMaybeTests.cs
+++ b/CSharpFunctionalExtensions.Tests/ResultTests/Extensions/AsMaybeTests.cs
@@ -1,0 +1,35 @@
+using Xunit;
+using FluentAssertions;
+
+namespace CSharpFunctionalExtensions.Tests.ResultTests.Extensions;
+
+public class AsMaybeTests : TestBase
+{
+    [Fact]
+    public void Error_returns_None()
+    {
+        var result = Result.Failure<T>(ErrorMessage);
+        result.AsMaybe().HasValue.Should().BeFalse();
+    }
+    
+    [Fact]
+    public void Success_returns_Some()
+    {
+        var result = Result.Success(T.Value);
+        var asMaybe = result.AsMaybe();
+        
+        asMaybe.HasValue.Should().BeTrue();
+        asMaybe.Value.Should().Be(T.Value);
+    }
+    
+    [Fact]
+    public void Null_value_returns_None()
+    {
+        var result = Result.Success<T>(null);
+        var asMaybe = result.AsMaybe();
+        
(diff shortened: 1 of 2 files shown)


RorySan:
Also, I've used TestBase instead of creating random types and data. Please confirm I'm using it properly.

bothzoli:
LGTM, as mentioned in #594 I think it'd make sense to add this extension to `UnitResult` as well.

RorySan:
Thank you @bothzoli , i'll work it out and update this. What about the other points i mention in the PR?

bothzoli:
Oh yeah, sorry:

1 - I'll pass on this one.
2 - I think you'll only need these if you do the `Task` implementations.
3 - Yeah, this is a bit tricky. On the one hand this makes sense from the `Maybe`-s perspective, yet it feels a bit weird that a `Success` turns into a `None`...
+1 - The usage of `TestBase` looks good to me here.

(I'm also merely another enthusiastic contributor to this repo, so I'm sorry I can't give more exact answers :))

RorySan:
Thanks for your input, may be @vkhorikov can help us out with task / value task. In the meantime i'll see about UnitResult and update the PR.

RorySan:
@bothzoli I've looked into `UnitResult` but by definition it _Represents the result of an operation that has no return value on success, or an error on failure._ I don't see how we can turn that into a Maybe when there is no value for the Some case. Can you clarify this?

bothzoli:
Ahh yeah, now that I think about it, you're perfectly right. By definition a `UnitResult` cannot have a value, so it cannot become a `Some` value 🤦
Sorry for making you run this circle in vain... 😔

vkhorikov:
Sorry for late reply.

1 -- Ideally, yes, you'd want to add overloads for both `Task` and `ValueTask`. I'll merge this PR as-is and you can add the overloads later.

2 -- Looks like some optimizations for nullable structs. No need to worry about it.

3 -- Yeah, that's a tricky one. Though I suppose since you (as a user) are already throwing away some of the information (i.e Result's error message), it's fine to throw away a little bit more (specifically, whether the result was successful or not)

PR #599, opened 2025-03-25
Add extensions: Bind optional, ensure not, map, optional, required

yringler:
Adds some more extension methods, mostly to make it easier to work with Result<Maybe

This is related to #545
There's probably a lot to go before these can get merged in.
Unit tests? Task and Value task?
But I figured might as well get the ball rolling.


vkhorikov:
Yea, tests + Task variations would be good to have too

